PR #21, opened 2015-05-15
rename typo in migration file name.

dzhuang:
typo error.


inducer:
Thanks! I was concerned that renaming migrations might make the migrations machinery unhappy, but it seems like it's OK.

PR #22, opened 2015-05-15
Adding i18n marks

dzhuang:
1. Add i18n marks to .py, .html(template) and .txt, with javascript in html files untouched.
1. Add migration file.


course/models.py
```diff
@@ -848,33 +860,34 @@ def __unicode__(self):
...
-            verbose_name='Kind')
+            verbose_name=_('Kind'))
     rule = YAMLField(blank=False, null=False,
-            verbose_name='Rule')
+            verbose_name=_('Rule'))
     active = models.BooleanField(default=True,
-            verbose_name=pgettext_lazy(
-                "Is the flow rule exception activated?", "Active"))
+            verbose_name=pgettext_lazy("is the flow rule exception activated?",
+                            "Active"))
```
inducer: is -> Is

course/templates/course/grade-notify.txt
```diff
@@ -1,15 +1,34 @@
...
   {{ page_title }}
 
-in {{ flow_session.flow_id }} of {{ course.identifier }}. The
+in {{ flow_id }} of {{ course_identifier }}. The
 full text of the feedback follows.
 
 -------------------------------------------------------------------
-{{ feedback_text|safe }}
+{{ feedback_text }}
 -------------------------------------------------------------------
```
inducer: Why remove the `|safe`?

relate/templates/bulk-result.html
```diff
@@ -1,13 +1,15 @@
 {% extends "base.html" %}
+{% load i18n %}
 
 {% load crispy_forms_tags %}
 
 {% block content %}
-<h2>{{ process_description }}</h2>
+<h2> {{ process_description }} </h2>
```
inducer: Why the extra spaces?

course/migrations/0060_add_verbose_name_and_lables_for_i18n.py
```diff
@@ -0,0 +1,746 @@
+# -*- coding: utf-8 -*-
```
inducer: Um, you've got a typo in your migration name. :)

`lable` -> `label`

inducer:
One more concern that's really up to you: Legally, you should probably add yourself to the copyright line at the top of the files you've touched.

course/models.py
dzhuang: It is context literal, like comments, which won't be show to users. Context are added because there're two "Active"s in the project, one refers to user status, another refers to the state of a exception rule.

course/templates/course/grade-notify.txt
dzhuang: transblock don't recognize filter and dot expressions, it is replace by alias "with feeback_text=feedback_text|safe" in the transblock tag.

course/models.py
dzhuang: I'll modify that anyway.

dzhuang:
There are some spaces and blank lines which are the results of wrong edit, I will check those later. For copyright issue, not for now, and thank you for your suggestion.

course/auth.py
```diff
@@ -119,15 +120,25 @@ def __init__(self, impersonator, *args, **kwargs):
...
-                    (u.id, "%s - %s, %s" % (u.email, u.last_name, u.first_name))
-                    for u in sorted(impersonees,
-                        key=lambda user: user.last_name.lower())
-                    ],
+                    # Translators: information displayed when select user
+                    # for impersonating, you can use your prefered
+                    # style. For example %(user_lastname)s,
+                    # %(user_firstname)s (user_email)s, but all three
+                    # strings should be used.
+                    (u.id, _("%(user_email)s - %(user_lastname)s, "
```
inducer: I would disagree with the statement that translators are free to choose their preferred style. There is a usability implication here. Leading with the email makes the field searchable by user's email addresses in most browsers. In addition, the field is sorted by a user's last name, making lookup by last name practical.

course/analytics.py
```diff
@@ -97,14 +99,29 @@ def add_data_point(self, value, weight=1):
...
         else:
             if (self.num_max_value is not None
                     and value > self.num_max_value):
-                self.add_data_point("(value greater than max)", weight)
+                self.add_data_point(
+                    "".join([
+                            "(",
+                            pgettext("Value of grade", "value greater than max"),
+                            ")"]),
+                    weight)
```
inducer: Why are the parens split off from the translated text?

inducer:
Thank you for the monumental amount of work you've put into this! I noticed a few minor niggles as I was reviewing it (as you saw above), but no showstoppers.

course/auth.py
inducer: Fixed in 152bbbe

dzhuang: In fact, I add that statement mainly for Chinese users, for convenience, as we always use full name with the format %(lastname)s%(firstname)s, without separator between the two string

course/analytics.py
dzhuang: Just don't want distract the translator from characters which don't require translation. Sometimes I forget to put parentheses in translated text, or spaces at the end of a string, and then it might result in a bad looking.

PR #24, opened 2015-06-02
escaping HTML characters in commit messages.

dzhuang:
If commit messages contain characters like "\>", "\<", "\&", or strings like "\<b\>", "\<p\>", the course content update page will not render correctly.

### DIFF
--- a/course/versioning.py
+++ b/course/versioning.py
@@ -50,6 +50,7 @@ from course.models import (
 from course.utils import course_view, render_course_page
 import paramiko
 import paramiko.client
+import cgi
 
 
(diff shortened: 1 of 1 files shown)


course/versioning.py
```diff
@@ -470,6 +471,9 @@ def update_course(pctx):
                     ugettext("Current preview git SHA"),
                     ":</b> ",
                     ugettext("None")]))
+        
+    thestring="".join("<p>%s</p>" % line for line in text_lines)
+    print thestring
```
inducer: This looks like leftovers from debugging.

inducer:
Thanks!

PR #26, opened 2015-06-09
fix i18n mark error.

dzhuang:


inducer:
Thanks for the patch!

Btw, if you have Chinese translations sitting around, I'd like to encourage you to submit them, so that others can potentially benefit from them.

dzhuang:
Actually I am considering to submit my translations. But at the moment, I found I don't understand some terms correctly, and there are still a few errors in marks. Should be done for a while.

PR #31, opened 2015-06-18
add zh-CN locale

dzhuang:
include simplified Chinese translation.


inducer:
Three small wrinkles:

* On line 69 of `django.po`, there is a blank translation.
* I don't think the binary compiled file `django.mo` should be included.
* Instead, the docs should tell the user to run `django-admin compilemessages`.

dzhuang:
Blank translation means "don't translate". I leave them untranslated because I can't find accurate and compact words to express the same idea in Chinese.

dzhuang:
updated.

inducer:
Thanks very much for your contribution! (And thanks for explaining!)

PR #35, opened 2015-07-23
fix i18n errors and update translations.

dzhuang:
* fix wrong marks in template ``flow-start.html``.
* update zh-CN translations.


inducer:
Thanks!

PR #36, opened 2015-07-23
only chunks which are .shown should be listed.

dzhuang:


inducer:
This filtering is actually done here:

https://github.com/inducer/relate/blob/master/course/content.py#L708

Did you encounter a situation where something slipped through?

dzhuang:
Yes. The object passed to  the template seems to be not filtered,  ``for chunk in course_desc.chunks``.

inducer:
Where? The course page view uses the function I pointed to above:

https://github.com/inducer/relate/blob/master/course/views.py#L132

dzhuang:
Sorry for my bad expressions. It's not about the contents rendered in course page, but the title hyperlink in the dropdown menu.

https://github.com/inducer/relate/blob/master/course/templates/course/course-page.html#L8

I now think my solutions is not so good, this line can be simply changed to:

```
{% for chunk in chunks %}
```

inducer:
Good catch, fixed in 120e619.

PR #37, opened 2015-07-23
fix: flow with 0 percentages displayed (none) in result in flow-start…

dzhuang:
flow with 0 percentages displayed ``(none)`` in result in ``flow-start.html``.


inducer:
Oops, good catch!

PR #38, opened 2015-07-24
localization of datetime string format.

dzhuang:
Need python package babel.


inducer:
I'd like to request a small change: The date formatting should be a separate function. Otherwise, (1) the return type of `local_now` changes based on a flag, and (2) if I've got another date to format, I'd have to reimplement that piece of `local_now`.

dzhuang:
Yes, you're right. I'll provide another solution later.

PR #39, opened 2015-07-26
Create a function "format_datetime_local" to localize datetime str.

dzhuang:
See http://babel.pocoo.org/docs/api/dates/#date-and-time-formatting for customizing the output format.


inducer:
Looks good. Thanks for the contribution!

PR #40, opened 2015-07-26
update i18n marks and translations of email text.

dzhuang:
Previous version reformatted the strings of email templates, which will result in bad indentations , bad line breaks and empty lines in emails. This PR fix those errors. However, those hints for translators do not appear in the ``django.po`` files generated. I have not idea how to fix it.


inducer:
Thanks! Merging this for now, we'll continue worrying about the hints...

PR #42, opened 2015-07-30
Add MultipleChoiceQuestion.

dzhuang:
Based on ``ChoiceQuestion``, add an optional attribute ``full_match``

Optional. ``True`` or ``False``. If true (default), only answers which fully matches the correct answer is correct.  If false, answers with subset of correct choices will get a proportional credit.

### DIFF
--- a/course/page/choice.py
+++ b/course/page/choice.py
@@ -45,6 +45,14 @@ class ChoiceAnswerForm(StyledForm):
         # Translators: "choice" in Choice Answer Form in a choice question.
         self.fields["choice"].label = _("Choice")
 
+class MultipleChoiceAnswerForm(StyledForm):
+    def __init__(self, field, *args, **kwargs):
+        super(MultipleChoiceAnswerForm, self).__init__(*args, **kwargs)
+
+        self.fields["choice"] = field
+        # Translators: "Multiple choice" in Choice Answer Form in a mutliple choice question.
+        self.fields["choice"].label = _("Multiple Choices")
+
 
 # {{{ choice question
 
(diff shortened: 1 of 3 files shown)


course/page/choice.py
```diff
@@ -232,7 +240,135 @@ def normalized_answer(self, page_context, page_data, answer_data):
         return self.process_choice_string(
                 page_context,
                 self.page_desc.choices[permutation[choice]])
+# }}}
+
+# {{{ multiple choice question
+
+class MultipleChoiceQuestion(ChoiceQuestion):
+    """
+    A page asking the participant to choose one of multiple answers.
```
inducer: That description doesn't quite match what I imagine this does.

course/page/choice.py
```diff
@@ -45,6 +45,14 @@ def __init__(self, field, *args, **kwargs):
         # Translators: "choice" in Choice Answer Form in a choice question.
         self.fields["choice"].label = _("Choice")
 
+class MultipleChoiceAnswerForm(StyledForm):
+    def __init__(self, field, *args, **kwargs):
+        super(MultipleChoiceAnswerForm, self).__init__(*args, **kwargs)
+
+        self.fields["choice"] = field
+        # Translators: "Multiple choice" in Choice Answer Form in a mutliple choice question.
+        self.fields["choice"].label = _("Multiple Choices")
```
inducer: I'd perhaps only call this "Choices".

inducer:
This looks great. I'll go ahead and fix the two niggles that I found as well as add a test to the test flow in the sample content.

dzhuang:
I forget to modify  the descriptions :-). 

BTW, I'm not sure whether other descriptions, along with the new attribute name ``full_match``, make sense in English. PLS correct those expressions, thx.

inducer:
I've made a bunch of changes in 79cd1ff: In particular, I've changed how partial credit scoring works and how it is enabled.

PR #45, opened 2015-08-08
update i18n.

dzhuang:
The following are updated:

* ``Translators:`` tags must be right above the literals to be translated.
* ``Translators:`` tags in email templates are moved to a new txt file for generating the help message in .po files.
* i18n for email templates are updated.
* all .po files are regenerated.
* zh_CN translations are updated.


inducer:
Thanks!

PR #46, opened 2015-08-11
Fix: allow calculable answers for FloatMatcher.

dzhuang:
Fix: for float type answer in ``FloatMatcher``, it does not allow ``value`` such as ``1/3``, or ``1+2``, we have to use ``0.333`` or ``3`` instead. And answer submitted is also restricted.


course/page/text.py
```diff
@@ -373,33 +374,60 @@ def __init__(self, vctx, location, matcher_desc):
...
-            float(s)
+            float(eval(s))
         except:
             tp, e, _ = sys.exc_info()
             raise forms.ValidationError("%(err_type)s: %(err_str)s"
                     % {"err_type": tp.__name__, "err_str": str(e)})
 
     def grade(self, s):
-        answer_float = float(s)
+        answer_float = float(eval(s))
```
inducer: This is a nonstarter, unfortunately, because this allows arbitrary code execution based on input into a web form.

inducer:
It may be possible to use `parse_sympy` in conjunction with sympy's `evalf` to this end:

https://github.com/dzhuang/relate/blob/float_matcher/course/page/text.py#L289

inducer:
Another stylistic point: I'd like to keep the conversion from string to float consistent, so instead of littering `eval` everywhere, there should be a central place that does the conversion that's called from all relevant spots.

course/page/text.py
dzhuang: You are right !  Thanks for the information.

dzhuang:
Thank's for the suggestions, and sorry for the bad coding.

dzhuang:
updated using parse_sympy and Sympy.evalf

inducer:
Looks good, thanks!

PR #48, opened 2015-08-17
Try to deploy MultipleTextQuestion.

dzhuang:
Mobile view need to be improved. Any suggestions will be appreciated.

### DIFF
--- a/course/flow.py
+++ b/course/flow.py
@@ -1100,6 +1100,8 @@ def view_flow_page(pctx, flow_session_id, ordinal):
                 # because we were allowed this far in by the check above
 
                 feedback = None
+                messages.add_message(request, messages.ERROR,
+                        _("Failed to submit answer."))
 
                 # continue at common flow page generation below
 
--- a/course/page/choice.py
+++ b/course/page/choice.py
@@ -137,8 +137,8 @@ class ChoiceQuestion(PageBaseWithTitle, PageBaseWithValue):
             raise ValidationError(
                     string_concat(
                         "%(location)s: ",
-                        "one or more correct answer(s) "
-                        "expected, %(n_correct)d found")
+                        _("one or more correct answer(s) "
+                        "expected, %(n_correct)d found"))
                     % {
                         'location': location,
                         'n_correct': correct_choice_count})
@@ -228,7 +228,7 @@ class ChoiceQuestion(PageBaseWithTitle, PageBaseWithValue):
 
     def correct_answer(self, page_context, page_data, answer_data, grade_data):
         corr_idx = self.unpermuted_correct_indices()[0]
-        return (_("A correct answer is: %s")
+        return (string_concat(_("A correct answer is"), ": '%s'.")
                 % self.process_choice_string(
                     page_context,
                     self.page_desc.choices[corr_idx]).lstrip())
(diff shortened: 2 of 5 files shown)


inducer:
Thank you for your contribution! This mostly looks extremely good, and it's largely ready to go, but there are a few wrinkles that I would like to talk through with you.

- Sort of least importantly, I have made a number of style fixes to your code. You can find the branch with those changes here: https://github.com/dzhuang/relate/pull/1 (kind of strangely, this is a sub-pull request to this pull request)
- `text.py` is getting kind of long. I was wondering how you might feel about splitting that file, in particular about splitting out there multiple inline question type.
- the multiple inline question type doesn't really provide great feedback, in the sense that it is hard to tell which partial answers are correct and which ones are wrong.
- if one of the fields in the multiple inline question type encounters a validation failure, the resulting validation error kind of messes up the page layout. That's a little unfortunate, but overall I think that is something I would be willing two live with.

dzhuang:
Try to use tooltip to show the error message.  The feedback problem will be a Todo.

inducer:
Uh-oh, it seems like we have overlapped in making changes to this code. Except for your last commit, everything is merged to master now. I have messed around with the JavaScript for the inline question page as well, and it seems to work as intended for me. What specific behavior were you trying to fix using your last commit?

At any rate, thank you again for another fine contribution to the code!

dzhuang:
I noticed the conflicts just now. Now they are resolved.

dzhuang:
I am happy to make improvement on this great project, and thank you for all your valuable tips.

inducer:
Hey, what you did there is simply undoing my changes and reverting back to your version. That is not quite what I have in mind :-). I changed the JavaScript so that the pop overs would activate on being focused, and I made the code that triggers the LaTeX rendering a little bit more intelligible (at least to me). What specifically were you changing, and how can we make the final code do what both of us wanted to accomplish?

dzhuang:
Sorry, I'll try to revert back to your version, as I didn't notice the what caused the conflicts.

dzhuang:
I had thought that the conflict is cause by the commits of mine, without knowing you've make revision in the code, sorry. The previous code on popup is to make sure at most one popover is activated. I haven't test your code, and I will try it 8 hours later. It's too late here (1:30am local time) and I can keep my brain running. That's one reason I did the stupid operation just now.

dzhuang:
One more thing, do you have any idea on the view of inlineMultipleQuestion on a mobile devices? It displays weird that the text and input field are broken into multiple lines. For example, a question: `` Foo and [[ bar ]]. `` one a mobile device will render a page with three lines, the last of with contains only a ``.``.

 ```
Foo and

the input field

.
  ```

inducer:
OK, once you've gotten some sleep, let me know if you would still like me to merge the remainder of this pull request.

Regarding the rendering on mobile devices, I unfortunate do not have the time to investigate this right now. But as you can see, I have left this as an open bug, so that we can get back to it later.

Regarding the data original title attributes, I am not sure I quite understand what you're trying to say. Could you please try and explain differently?

dzhuang:
About the ``data-original-title`` attribute: By default, both popover and tooltip read from that element attribute to set their title. Thus, if an element have a popover and the attribute is given, when a tooltip is added to the element, its title will be the same as that of the popover. Now I realized that was not the case. We can use javascript to solve them problem. 

Reference: http://stackoverflow.com/questions/24107002/bootstrap-3-popover-and-tooltip-on-the-same-element

inducer:
Looks good. Thanks again for your contribution!

PR #50, opened 2015-08-21
Inline multiple form

dzhuang:
- improve feedback of ``InlineMultipleQuestion``.
- Field validation of field, especially for text type question (``ShortAnswer``).

### DIFF
--- a/course/page/inline.py
+++ b/course/page/inline.py
@@ -109,20 +109,21 @@ class InlineMultiQuestionForm(StyledInlineForm):
             field_name_idx = instance_idx.name
             if hasattr(instance_idx, "matchers"):
                 for validator in instance_idx.matchers:
-                    try:
-                        validator.validate(cleaned_data[answer],
-                                validate_only=True)
-                    except:
-                        from traceback import print_exc
-                        print_exc()
+                    if answer in cleaned_data:
+                        try:
+                            validator.validate(cleaned_data[answer],
+                                    validate_only=True)
+                        except:
+                            from traceback import print_exc
+                            print_exc()
 
-                        import sys
-                        tp, e, _ = sys.exc_info()
+                            import sys
+                            tp, e, _ = sys.exc_info()
 
-                        self.add_error(field_name_idx,
-                                       "%(err_type)s: %(err_str)s" % {
-                                            "err_type": tp.__name__,
-                                            "err_str": str(e)})
+                            self.add_error(field_name_idx,
+                                           "%(err_type)s: %(err_str)s" % {
+                                                "err_type": tp.__name__,
+                                                "err_str": str(e)})
 
 
 def get_question_class(location, q_type, answers_desc):
--- a/course/page/inline.py
+++ b/course/page/inline.py
@@ -111,19 +111,12 @@ class InlineMultiQuestionForm(StyledInlineForm):
                 for validator in instance_idx.matchers:
                     if answer in cleaned_data:
                         try:
-                            validator.validate(cleaned_data[answer],
-                                    validate_only=True)
+                            validator.validate(cleaned_data[answer])
                         except:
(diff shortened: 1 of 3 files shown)


course/flow.py
```diff
@@ -1046,25 +1046,58 @@ def view_flow_page(pctx, flow_session_id, ordinal):
                     or flow_permission.change_answer
                     in permissions)
 
+            # {{{ duplicate code for make_feedback_form
```
dzhuang: I introduced a special method to generate in-form feedback for InlineMultipleQuestion, to show which answer is correct and which is not, and here I used duplicate code (NOT DRY), because I find it huge work to reorganize the logic.

inducer:
Thanks for your contribution! A few comments on this patch. First, I've sent you a request that addresses the underlying architectural issue that made the duplicated code necessary. It's somewhat major surgery, so it would be great if you could test what I sent as thoroughly as you can.

Next, a few comments on this code in no particular order.

* The tooltip code and the rendering of the LaTeX clash with each other a little bit, in that once the tooltip is placed, some latex may get rendered and change the with of the preceding text. This may move the input field with which the tooltip was associated, but it seems as though the tooltip does not move itself in response to that. I think the message from this is that latex should probably be used sparingly in this type of question, I'm not sure what else we could do about it.

* I'm not sure I completely understand what you intend with the `validate_only` keyword argument, why is that necessary? Can you please explain?

* Something about the rendering of validation outcomes into tooltips seems to break Django's own validation handling. If I enter non-numerical data in the test multiple question in [the sample content](/inducer/relate-sample), and then press "Save answer" (but not "submit final answer") a message along the lines of "KeyError: 'blank_2' appears underneath the form field in addition to the tooltip.

dzhuang:
For your third comment, that's because the javascript code in [lines](https://github.com/inducer/relate/blob/master/course/templates/course/custom-crispy-inline-form.html#L29-L47) consider only the first error. While there are more errors for the same field, extra errors will be left there untouched. 

BTW, the "KeyError: 'blank_2'"  error seems to be a bug with question which provide both float type answer and other type answers, and I've commented it in your [sample content](https://github.com/inducer/relate-sample/commit/3e9055083e4a40feb2a4a458ee8b52d7949eb2b4#diff-91cc9da1965d8ddb06467443c97aab15R471).

dzhuang:
For your fist comment, is it possible to render the tooltips after all the mathjax objects finished typesetting? I see [this](https://github.com/mathjax/mathjax-docs/wiki/Event-when-typesetting-is-done%3F-Rescaling-after-rendering...) in official document.

dzhuang:
For you second comment, about the ``validate_only``, maybe I've been wrong in doing so. My idea is that, those validate functions are originally designed for questions with only one field, where the error message is pointed to that unique field. However, for questions with multiple fields, an shown error message should correspond to the field which have the error. But the solution I provided may not be DRY.

dzhuang:
Thank you for your surgery to the project. That must be a time consuming work. I think there are more thing can be done after that improvement.

dzhuang:
The last PR should fix the issues related to your first and third comments. For the necessity of ``validate_only``, I have no idea right now.

dzhuang:
I find where the bug is. In [line111](https://github.com/inducer/relate/pull/50/files#diff-7b95f7e2c3ef72b907eb564186ec1552R111), if a field failed to pass validation, it's key will be popped from the cleaned_data dict. However, if the ``floatMatcher`` is not the final validator, the validation loop is going to validate that non-existing data, that will raise a KeyError of the field. 

That might also be the reason why when a TextQuestion contain's a float type answer, other answers (including another float type answer) will be suggested as correct.

inducer:
Great job. Thanks again for making these improvements happen, and for being responsive to comments.

Issue #52, opened 2015-08-22
TextQuestion with float type answer will reject all other answer types.


dzhuang:
When a TextQuestion contains a float type answer, other type answer will be suggested as incorrect.

dzhuang:
Thank you.

PR #53, opened 2015-08-23
Fix in accordance with fix of issue #52.

dzhuang:
Update text question validation in accordance with https://github.com/inducer/relate/commit/32c3fecac797e4bcd9866e020eab5a00460d098e , along with minor modifications on docs.


inducer:
Thanks!

PR #55, opened 2015-08-24
try to re-organize home page.

dzhuang:
In a simplest way, leaving permission of access untouched.

### DIFF
--- a/course/admin.py
+++ b/course/admin.py
@@ -160,8 +160,8 @@ class CourseAdminForm(forms.ModelForm):
 
 
 class CourseAdmin(admin.ModelAdmin):
-    list_display = ("identifier", "hidden", "valid", "listed", "accepts_enrollment")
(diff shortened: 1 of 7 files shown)


inducer:
Thanks for getting started on this! As I said, I'm not a fan of the policy decision to hide ended courses. In addition, one thing that I would like to happen as part of this is that courses are sorted by their start/end date or semester in some fashion, to sort of unjumble the homepage a bit. I'll see when I have time to take a look.

dzhuang:
The main purpose of this PR is to prevent student from trying to enroll ended courses (intentionally or messed up by the layout of home page), and participants can easy find ended courses they've attended. However, I must admit that the tweak of this PR is not considerate enough. 

Your idea about using start/end date along with sorting will be better than mine. I'll wait for your improvement. Thank you!

PR #56, opened 2015-08-25
User name validation, reset password error, i18n update.

dzhuang:
1. Currently, user name allow any kind of string, including spaces and non-ascii characters. However, that will cause errors in Admin: ``Enter a valid username. This value may contain only letters, numbers and @/./+/-/_ characters.`` The PR use django default validators to fix the problem, from [line 291](https://github.com/inducer/relate/pull/56/files#diff-c2451d72ff86377e6511c9368f632f0bR291). What's more, some literals changed to those of django default, so that stock translations are provided for most languages for those literals (see [here](https://github.com/django/django/tree/8047e3666b0b50bb04e6f16c2a4fb21ddfd5713f/django/contrib/auth/locale)).

1. Fix bug on [line 404](https://github.com/inducer/relate/compare/inducer:master...dzhuang:user_name_validation?expand=1#diff-c2451d72ff86377e6511c9368f632f0bL404) when providing a non-exist email for resetting password, which will cause errors.

1. Update i18n for new literals.

### DIFF
--- a/course/auth.py
+++ b/course/auth.py
@@ -285,14 +285,12 @@ def sign_in_by_user_pw(request):
 class SignUpForm(StyledModelForm):
     username = forms.CharField(required=True, max_length=30,
             label=_("Username"),
-            help_text=_("Required. 30 characters or fewer. Letters, "
-                "digits and @/./+/-/_ only."),
             validators=[
                 validators.RegexValidator('^[\\w.@+-]+$',
                     string_concat(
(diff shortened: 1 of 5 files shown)


inducer:
Line **404** has a bug about something not existing? Thanks for making my morning. :)

inducer:
Looks good. Thanks for the fixes!

PR #68, opened 2015-09-08
update gopp.name only when a new flow session is created

dzhuang:
Drawback is that gopp.name will only be updated when a new session of that flow is started.


inducer:
You're of course right with that drawback of your solution, but at the same time the situation is arguably better than it was with the patch included. Once again, thanks for your contribution!

PR #75, opened 2015-09-11
Warning a non-existing events_file which is not named "events.yml".

dzhuang:
- If Course.events_file  is "foo.yml" (not "events.yml"), and if "foo.yml" doesn't exist in the course repository, add warnings in validation.
 - i18n error fixes.


inducer:
Looks good. Thanks!

PR #85, opened 2015-09-15
when All questions are saved or submitted, add a quick button for sub…

dzhuang:
When answers for *All* questions are saved or submitted, add a quick button for submitting session.

### DIFF
--- a/course/flow.py
+++ b/course/flow.py
@@ -295,9 +295,7 @@ def count_answered(fctx, flow_session, answer_visits, force_gradeable=True):
     return (answered_count, unanswered_count)
 
 
-def count_answered_gradable(fctx, flow_session, answer_visits,
-                           force_gradeable=True):
-
+def count_answered_gradable(fctx, flow_session, answer_visits):
     return count_answered(fctx, flow_session, answer_visits,
                            force_gradeable=True)
 
--- a/course/flow.py
+++ b/course/flow.py
@@ -262,7 +262,7 @@ def get_interaction_kind(fctx, flow_session):
     return ikind
 
 
-def count_answered_gradable(fctx, flow_session, answer_visits):
+def count_answered(fctx, flow_session, answer_visits, force_gradeable=True):
     all_page_data = (FlowPageData.objects
             .filter(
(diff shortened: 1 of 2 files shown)


inducer:
Two concerns:

* first, and more importantly, there is a situation where you have provided answers for all questions, but do not yet want to submit your flow session. Imagine the flow consisting of all code questions, which allow many submissions, until the participant is satisfied with the result of their computation. In this situation, I could imagine that a participant has provided answers for each question, but he or she is not finished working on them. If I understand this patch correctly, it will put a blinking button in their face all the time. Not ideal…

* Secondly, the function that you call (`count_answered`) is quite expensive in terms of database access and Python computation. I know I added a similar call in 933822064 (and I feel bad about that), but `assemble_answer_visits` is on a whole another level in terms of computational expense. I fear that this may significantly impact response times for flow pages, which is where I expect people to spend most of their time interact with relate.

I support the general notion though, reminding people to submit their sessions is probably good idea.

dzhuang:
You are right !!

How about this? Warning messages at the top of course home page (rather than flow pages) reminding student that he/she has a new task or unfinished session, with a link that he/she can navigate to see a summary of those sessions.

inducer:
I am liking half of that idea. :-)

* The list of unfinished sessions should be easy. (And it should be reasonably cheap to assemble.) I would realize that not necessarily is something that automatically pops up at the top of the course page, but as a predefined Jinja macro. Then each instructor is at liberty to include it or leave it out. (In my planning, the course calendar is headed in the same direction.) If you could cook up a patch for this, I would be grateful.

* The list of new tasks on the other hand is much much trickier. Not every new flow that a student may have access to is automatically any task that they are asked to complete. As such, this feature would need infrastructure to enumerate them. And once you've built that, you will need to ask instructors to tag (somehow) a subset of their flows as new tasks. That sounds irritating and not very useful, especially when compared to the much more straightforward approach of simply manually putting a list of flows in the right spot on the course page.

dzhuang:
I'll have a try when I have time. Let's close this PR now.

PR #89, opened 2015-09-17
fix i18n mark errors.

dzhuang:


locale/de/LC_MESSAGES/django.po
```diff
@@ -3894,15 +3907,33 @@ msgid ""
 "If this was not you, it is safe to disregard this email.\n"
 msgstr ""
 
-#: .\course\utils.py:337
+#: .\course\templates\course\task-monitor.html:12
+#, fuzzy
```
dzhuang: Just for your reference, you need to find all blocks with ``#, fuzzy`` lines, remove those lines and adjust your translations, when the django.po files are modified by ``makemessages`` command. Or those literals won't be translated.

course/views.py
```diff
@@ -1041,8 +1041,9 @@ def monitor_task(request, task_id):
             progress_percent = 100 * (current / total)
 
         progress_statement = (
-                _("%d out of %d items processed.")
-                % (current, total))
+                _("%(current)d out of %(total)d items processed.")
```
dzhuang: Just for your reference, if more than 1 string is passed, we need to use this ``dict`` format to map them, or else ``makemessages`` will raise errors.

inducer: Just learned that. :) Thanks for the pointer.

inducer:
Argh. Sorry. I think I caused a conflict here. I manually transferred your code changes. Could you fix up the Chinese `.po` file?

dzhuang:
It's ok. I think since you start translate de, I will not update the django.po file of de in the future. Because I am using Windows platform, which will generate different format of .po files, for example, slash and backslash, I think that may cause conflicts frequently.

inducer:
Thanks!

Issue #91, opened 2015-09-18
If the enrollment preapprovements are added after the enrollment requests of students


dzhuang:
Then the students will still need to wait for the approvement of the tutor.

inducer:
Easy patch here:

https://github.com/inducer/relate/blob/16fe41435c85c413dc97db64e631f046e45f93b3/course/enrollment.py#L261

dzhuang:
Thank you, I tried to fix it in https://github.com/inducer/relate/pull/94.

Issue #90, opened 2015-09-17
Don't send emails to course staff in language of the student's browser


inducer:
@inducer doesn't read Chinese.

dzhuang:
Opps.... Seems there are Chinese students using your instance :-).

PR #94, opened 2015-09-20
process pending requests included in preapprovement.

dzhuang:
Try to fix https://github.com/inducer/relate/issues/91.

### DIFF
--- a/course/enrollment.py
+++ b/course/enrollment.py
@@ -138,7 +138,8 @@ def enroll(request, course_identifier):
                 "user": user,
                 "course": course,
                 "admin_uri": request.build_absolute_uri(
-                        reverse("admin:course_participation_changelist"))
+                        reverse("admin:course_participation_changelist")
+                                + "?status__exact=requested")
                 })
 
             from django.core.mail import send_mail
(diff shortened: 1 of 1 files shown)


inducer:
Looks good, thanks!

course/enrollment.py
```diff
@@ -178,6 +179,15 @@ def decide_enrollment(approved, modeladmin, request, queryset):
...
+        send_enrollment_decision(participation, approved, request)
+
+        count += 1
+
+    messages.add_message(request, messages.INFO,
+            # Translators: how many enroll requests have ben processed.
+            _("%d requests processed.") % count)
+
+def send_enrollment_decision(participation, approved, request):
         with translation.override(settings.RELATE_ADMIN_EMAIL_LOCALE):
```
dzhuang: According to https://github.com/inducer/relate/issues/90, seems it's not a problem to use which locale when sending email to student, I think it is a feature that a student receive notifications using his/her prefered language.

inducer: Yes. This message is cc'd to course staff though, and it's initiated from an HTTP request coming from a staff account, so the default locale seems like a fair fallback. The way I view it, both staff and students should be able to understand the default locale...

PR #98, opened 2015-09-21
improve ui.

dzhuang:
1. Improve display of faketime and impersonation messages, add shortcut disable button on top of base page.
 1. Open staff menu items in new tab.
 1. Add icons for different type of messages.

### DIFF
--- a/course/auth.py
+++ b/course/auth.py
@@ -163,8 +163,6 @@ def impersonate(request):
         if form.is_valid():
             user = get_user_model().objects.get(id=form.cleaned_data["user"])
 
-            messages.add_message(request, messages.INFO,
-                    _("Now impersonating '%s'.") % user.username)
             request.session['impersonate_id'] = user.id
 
             # Because we'll likely no longer have access to this page.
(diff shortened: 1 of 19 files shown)


inducer:
I like most of this, except for all the commands opening new tabs. What was the rationale for that?

dzhuang:
I think the new tab command is helpful especially for debugging. When I am faking datetime,  impersonating, updating course materials, I'll need to navigate back at least 3 pages before I can resume to the page before the faking or impersonating.

inducer:
Okay, so how about this as a compromise: I can see the justification for forcing the opening of the link in a new tab for the debugging commands (set fake time, impersonate). For everything else, the justification seems a bit more shaky. That is especially true given that you can always get the same effect by middle clicking the menu link, and I like retaining that element of choice for the users.

dzhuang:
ok.
2015年9月21日 下午11:48，"Andreas Klöckner" <notifications@github.com>写道：

inducer:
Looks good, thanks!

PR #115, opened 2015-10-05
Chunk hack for jinja macro

dzhuang:
Work only for contents in chunks.

Jinja2 macro in repo file ``macros.jinja``.

```
{% macro show_if_after(datespec) -%}
{%- if now is defined and parse_date_spec(datespec) > now %}
<div markdown="block" class="hidden">
{%- elif now is defined and parse_date_spec(datespec) <= now %}
<div markdown="block" >
{% endif %}
{{ caller() }}
</div>
{%- endmacro %}
```

The macro should be imported with `` with context``, i.e.:
```
{% from "macros.jinja" import show_if_after with context %}
```


inducer:
As you can probably imagine, I'm not super-excited about this patch:

* First, there's no real reason why Jinja in page chunks should behave differently from the way they do everywhere else.
* Next, having to hand-write/import that macro every time seems kind of clunky.
* Lastly, this ignores caching and therefore will show the wrong content at times.

I'll leave this open, as I think there are pieces that are worth keeping.

dzhuang:
Yes, you are right. Maybe the value of this PR is just to show what we can do if we can pass some markup context to the content rendered.

PR #117, opened 2015-10-06
Add Select2 widget for updating course SHA

bowmanat:


inducer:
Looks great! Thanks for the patch.

PR #120, opened 2015-10-09
Add option to preview off of preview

bowmanat:
It would be nice to switch directly between previews, rather than pressing "End Preview" and then "Preview", since I often need to switch from preview to preview when debugging homework questions.  

Unless I missed something subtle, this seems to be a simple adjustment.  This patch worked fine in my local testing.


inducer:

PR #124, opened 2015-10-17
Prune remote refs when fetching

bowmanat:
As discussed in a previous issue, this will keep the SHA dropdown from being cluttered with old branches.


inducer:

PR #125, opened 2015-10-20
prevent two sessions in close succession, time interval = 5 sec.

dzhuang:
Works when JS is disabled, and when the student is not intended to multiple post the requests.

### DIFF
--- a/relate/settings.py
+++ b/relate/settings.py
@@ -164,6 +164,8 @@ RELATE_CACHE_MAX_BYTES = 32768
 
 RELATE_ADMIN_EMAIL_LOCALE = "en_US"
 
+RELATE_SESSION_RESTART_COOLDOWN_SECONDS = 10
+
 for name, val in local_settings.items():
     if not name.startswith("_"):
         globals()[name] = val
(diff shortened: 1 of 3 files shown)


inducer:
Thanks for coding that up and submitting it!

To be properly race-free (i.e. protected by a `serializable`-level transaction), the query for `past_sessions` should be within `post_start_flow`. I also wonder if we should make that value (5 seconds) configurable, such as by: `getattr(settings, "RELATE_SESSION_RESTART_COOLDOWN_SECONDS", 10)`.

dzhuang:
Thank you for your suggestions. I've update the code as suggested.

relate/settings.py
```diff
@@ -164,6 +164,8 @@
 
 RELATE_ADMIN_EMAIL_LOCALE = "en_US"
 
+RELATE_SESSION_RESTART_COOLDOWN_SECONDS = 10
```
inducer: Because we're using `getattr()` with a default, this isn't needed. I've gone ahead and removed it.

inducer:
Looks good. Thanks!

PR #141, opened 2016-01-11
Fix prevent-two-sessions-in-close-succession when faking time.

dzhuang:


inducer:
Thanks for your contribution!

PR #143, opened 2016-01-13
I18n updated

dzhuang:
Updated i18n tags and zh_CN translation. Fixed markdown in model help_text.


inducer:
Thanks for the updates!

PR #145, opened 2016-01-17
Fix unique field blank value duplication error in admin.

dzhuang:
https://code.djangoproject.com/ticket/4136#comment:33

That is also the case for ``sign_in_key`` field, but that field is not displayed in admin.


inducer:
Thanks! Added a comment linking to the Django bug, since it's hard to decipher why that code is there if you don't know what the issue is.

PR #147, opened 2016-01-19
Preapproval by inst_id

dzhuang:


inducer:
I like the idea. That said, the migration needs to also convert over existing preapprovals to the new schema. Otherwise, I'd lose a bunch of preapproval data if I just ran this on the Illinois instance.

Here's an example of how to write a migration like that:

https://github.com/inducer/relate/blob/master/accounts/migrations/0007_migrate_user_status_data.py

inducer:
Never mind, I take that back.

inducer:
Thanks! Looks good to me.

dzhuang:
Thanks, seems it will do no harm to exist preapproval data.

Issue #146, opened 2016-01-19
Process pending request of course enrollment on course attribute change


dzhuang:
There are some situations, for example, when the tutor change the requirement of  ``enrollment_required_email_suffix`` from "@what.com" to "", all pending requests should be accepted. Is there a way to automatic the process?

inducer:
We should just attach some signal handlers to the relevant  spots and handle enrollment approval whenever the underlying data changes.
https://docs.djangoproject.com/en/1.9/topics/signals/

PR #148, opened 2016-01-19
Add a "retry enroll" button for qualified user to enroll again.

dzhuang:
Situations for this PR: 

1. A user had tried to enroll courses which require institutional_id, while he didn't set that value in his profile. He see in the course "Your enrollment request is pending. You will be notified once it has been acted upon." After his profile is updated, however, he has no way to apply for the enrollment, as he still see "Your enrollment request is pending...".

2. A user who has a non-verified institutional_id in his profile, he had applied for enrollment of the course, pending. After his institutional id is verified, the status of this request is still pending.

The solution to this issue (similar with https://github.com/inducer/relate/issues/146) in this PR is to determine whether the user is qualified for the course **when he visit course_page again**, after changes were made to his profile, or course configurations are changed (e.g., preapprovement). If yes, he will get a button on course_page to "retry enrollment".

### DIFF
--- a/course/templates/course/course-page.html
+++ b/course/templates/course/course-page.html
@@ -29,5 +29,14 @@
       </div>
     {% endif %}
 
-  {% endif %}
+    {% if show_retry_enroll_button %}
+          <div class="well">
+            <form method="POST" action="{% url "relate-enroll" course.identifier %}">
+              {% csrf_token %}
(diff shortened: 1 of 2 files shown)


inducer:
I'm not a huge fan--this seems clunky. I'd much rather solve #146 in a principled way. Let's discuss there. Closing this for now,  until we decide we still need this.

PR #149, opened 2016-01-19
Add inst_id field in user-profile-form.html

dzhuang:
Let users submit their own institutional ID. Once submitted, the field is disabled. Allowing no_id option, for other teachers/staffs who are not students, or in cases when some  students may be inclined to submit the inst_id later.


inducer:
Why  does this need to use AJAX?

dzhuang:
When a valid inst_id is submitted, I need to update the widget of inst_id and disable input of that field. By default, after submission of the form it will not redirect to other page, and I think ajax can do the job without refreshing the page.

inducer:
I see why you're doing it, but I'd still say AJAX is kind of too much complexity here--just make a new form instance after submission.

dzhuang:
Opps..., Thank you!

dzhuang:
Thank you for your precious suggestions, AJAX is removed. I tried to add a button on course page for student who didn't set up inst_id in his/her profile, and let him/her redirect back after updating the profile. What I am not sure are the English literals I added.

inducer:
This looks much better! I do have one more question though: I thought that the institutional ID input would be enabled until the ID is verified, and then disabled. (Same for the name.) Obviously, this verification would need to be done by an admin. I've tweaked the user admin in 895c6ca to make such admin verification plausible. I feel that with your current policy as given in `enable_inst_id`, we might as well leave the input enabled full-time, because once someone really *wants* to enter the wrong ID for themselves, they can, at any time. They just need to look at the RELATE source and see that all it takes is the extra GET parameter.

dzhuang:
The tweaks on widget (and the js as well) are just for better display the form, but as https://github.com/dzhuang/relate/blob/inst_id_in_profile_form/course/auth.py#L714-L719, once the ID is confirmed, it can only be modified by admin.

Maybe we can add some RELATE constant in settings to meet different needs, for example to disabled input immediately or after verification.

For me, it's hard to verify name, many of my students just leave them blank. With the inst_id, I am thinking of import a name list csv including inst_id and real name to map the account and the student. 
I currently have no idea of how to "verify" a students name and institutional ID. For me, maybe I will ask them to upload/email a scanned picture of their ID card, or photo, or face to face? That may be very time consuming. If there is no efficient way to  do the verification, maybe I'll set up the course with out the verification requirements on id at the beginning.

dzhuang:
I think your suggestion is more reasonalble . I will try to add a
RELATE_DISABLE_INSTITUTION_ID or something in localsettings.py to control
the behavior.
2016年1月22日 上午12:03，"Andreas Klöckner" <notifications@github.com>写道：

inducer:
Sounds good, will wait for updated PR.

dzhuang:
Here comes a problem, if the institutional ids and names are going to be verified, who are responsible for verifying them? It is important before more things can be done. 

Considering the following situation, there're 100 independent courses running on your instance, there maybe 5000 students applying for the 100 courses. So there will be 5000 new id need to be verified. Now the 100 instructor are responsible for the verification, how will each instructor know which subset of students he need to verify?

inducer:
Hey, I thought of that already. ;) 895c6ca lets you filter the user list in the admin by course participations. (even if they're just requests)

dzhuang:
Sorry, I didn't look into that last night as I was using mobile phone. So you mean that the verification will happen if only the student has requested for enrollment, and the verification should be done by the instructor of the course?

inducer:
That was the best idea I had--but I'm open to different ones.

dzhuang:
added RELATE_EDITABLE_INST_ID_PRE_VRF = True

inducer:
Some fixes in my `inst_id_in_profile_form` branch.

course/auth.py
```diff
@@ -680,9 +774,35 @@ def user_profile(request):
 
     user_form = None
 
+    def has_verified_inst_id(user):
```
inducer: Why does this ever return True for a non-verified ID?

course/auth.py
```diff
@@ -680,9 +774,35 @@ def user_profile(request):
...
+        if not EDITABLE_INST_ID_PRE_VRF:
+            return True and user.institutional_id or False
+        else:
+            if user.institutional_id_verified:
+                return True
+            else:
+                return False
+
+    def enable_inst_input(request):
```
inducer: Why allow editing institutional IDs just based on whether a GET parameter is present? If you want to be able to trust those inst IDs, then this makes no sense.

course/auth.py
dzhuang: I think when I don't want to verify user's student's id, that method should be named ``has_inst_id``. I should have make this clean or introduce another method.

course/auth.py
dzhuang: Yes. In fact, that is interface matter, just to make sure when users are login for the first time or if they are asked to update their inst_id in profile, all the fields are shown, if he/she is allow to modify those fields. This just skips a step for the user to uncheck the ``I have no inst_id`` checkbox. 

Do you suggest I should let the js to do the trick rather than in python backend?

dzhuang:
I will try to reorganize the code to make the logic better and reduce confusion.

course/auth.py
inducer: My vote would be for making sure the meaning of all the code here is crystal-clear and completely obvious from its naming. This code is sensitive in the sense that if it mishandles student info, then that's either a [FERPA](https://en.wikipedia.org/wiki/Family_Educational_Rights_and_Privacy_Act) violation (on my end) or an opportunity for cheating by students, or both.

course/auth.py
inducer: It's fine to do it on the server end as long as the naming makes it clear that it's not intended to be about rule enforcement logic but rather just UI convenience.

dzhuang:
Sorry, I just saw your last comments on this PR about your fixes.

dzhuang:
I've reorganized the code, and I tried to modify the namings of variables so that it won't cause confusion, but I am not confident it meet your requirements :). Any suggestions?

inducer:
Thanks for your contribution!

dzhuang:
Thanks for all your great suggestions, especially those beyond code it self.

PR #150, opened 2016-01-20
User model save trigger enrollment update.

dzhuang:
``send_enrollment_decision`` method is changed because model signals have no request objects. Thus sites framework should be enabled as an INSTALLED_APP (with migrate and update needed), or provide a  ``RELATE_SITE_DOMAIN`` in the settings.

TODO: update enrollment on course model saved signals.


dzhuang:
Updated and tested.

course/receivers.py
```diff
@@ -0,0 +1,109 @@
...
+    
+    if isinstance(instance, Course):
+        course = instance
+        requested_qset = Participation.objects.filter(
+                course=course, status=participation_status.requested)
+
+    elif isinstance(instance, User):
+        user = instance
+        requested_qset = Participation.objects.filter(
+                user=user, status=participation_status.requested)
```
inducer: There should be an else branch here to catch if `instance` is not of the expected type.

course/receivers.py
```diff
@@ -0,0 +1,109 @@
...
+
+    if not created:
+        if requested_qset is None:
+            pass
```
inducer: This looks fishy. Why is this here?

course/receivers.py
```diff
@@ -0,0 +1,109 @@
...
+
+        for requested in requested_qset:
+            if isinstance(instance, Course):
+                user = requested.user
+            elif isinstance(instance, User):
+                course = requested.course
```
inducer: Same.

course/receivers.py
```diff
@@ -0,0 +1,109 @@
...
+                        preapproval = (
+                             ParticipationPreapproval.objects.get(
+                                 course=course,
+                                 institutional_id__iexact\
+                                         =user.institutional_id))
+                    except ParticipationPreapproval.DoesNotExist:
+                        pass
+        pass
+
+    def enroll(status, role):
```
inducer: Could enroll be moved outside of this function?

local_settings.py.example
```diff
@@ -8,6 +8,15 @@ ALLOWED_HOSTS = [
...
 
+# Comment or leave the following line as is if you can run the
+# following command:
+#        python manage.py migrate sites
+# If that works without problem, edit in admin the "sites" block,
+# update the first/only record to your domain.
+# Otherwise, replace the following string with your own domain.
+
+RELATE_SITE_DOMAIN = "www.example.com"
+
```
inducer: I don't get the comment.

inducer:
It looks like this introduces usage of the Django sites framework. For what, when nothing else in Relate is sites-aware?

course/receivers.py
dzhuang: Thanks for pointing out.

course/receivers.py
dzhuang: Thanks, ``pass`` should be ``return``.

course/receivers.py
dzhuang: You're right, ``if created`` should be put at top to avoid unnecessary data base operation, and ``pass`` should be ``return``.

course/receivers.py
dzhuang: In fact this is copied from ``enrollment.py``, where there're two ``enroll`` methods, with one inside another. I just copied the embedded one. I will try to move the embedded one out, and import it here to avoid duplication.

dzhuang:
Yes, that's too expensive just for correct url in emails, and only those triggered by signals. I'll abandon this idea, and modify in the way you did, like [this](https://github.com/inducer/relate/blob/master/local_settings.py.example#L167).

local_settings.py.example
dzhuang: Sorry for my bad expression. My idea is that,  if sites can be properly migrated, then use ``Site_ID=1`` or something to generate the full url. 

Anyway, I've abandon that idea, use BASE_URL instead.

course/receivers.py
inducer: This didn't get fixed yet.

relate/settings.py
```diff
@@ -277,4 +277,9 @@
 
 # }}}
 
+# This makes sure the RELATE_BASE_URL is configured.
+assert local_settings["RELATE_BASE_URL"]
+print "The base url of your site will be %s" % (
```
inducer: Not a fan of printing from the settings.

inducer:
Couple suggestions:

* Rename `enroll` -> `enroll_view`.
* Rename `do_enroll` -> `handle_enrollment_request`. (and make it also handle rejections)
* Call `send_enrollment_decision` from `handle_enrollment_request`. (since they seem to be called together anyhow)

(this reduces API surface area and isolates "business logic" in one spot)

dzhuang:
Thank you for your patience,  for all the opportunities and suggestions.

course/receivers.py
dzhuang: Do you mean the ``pass`` issue? it's fixed while it displays a wrong version. See https://github.com/inducer/relate/pull/150/files#diff-4ae70464fc8d947d63bd94162a09979bR60.

inducer:
Cool. This looks good now. Thanks for the contribution, and for putting up with my nagging. :)

PR #153, opened 2016-01-27
Centralizing name formatting.

dzhuang:
A quick tweak.


inducer:
Not a huge fan. I think this should depend on locale rather than a user's name. But I could be talked out of that. Centralizing name formatting is a good idea though.

dzhuang:
Maybe done by [django format l10n](https://docs.djangoproject.com/es/1.9/topics/i18n/formatting/) is better, however, it also has its drawbacks. For example, there's an Indonesian student in my class, whose name is "Foo Bar". Using my locale, it will get a full name "FooBar", that will be awkward when the system send him email with that name.

dzhuang:
Adding three configurations: 
* ``RELATE_USER_FULL_NAME_FORMAT_METHOD``
* ``RELATE_EMAIL_APPELATION_PRIORITY_LIST`` and
* ``RELATE_DISPLAY_USER_FULLNAME_IN_LIST``

course/admin.py
```diff
@@ -591,8 +591,7 @@ def get_participant(self, obj):
     search_fields = (
             "flow_id",
             "participation__user__username",
-            "participation__user__first_name",
-            "participation__user__last_name",
+            "participation__user__full_name",
```
inducer: Disagreed here. This'll either not work or make the search very expensive--because the database can't carry it out on its own.

course/templates/course/participation-table.html
```diff
@@ -3,8 +3,12 @@
 <table class="table table-striped gradebook-participants">
   <thead>
     <th class="datacol">{% trans "User ID" %}</th>
+    {% if RELATE_DISPLAY_USER_FULLNAME_IN_LIST %}
+    <th class="datacol">{% trans "Full Name" %}</th>
```
inducer: Why not just "Name" at this point?

local_settings.py.example
```diff
@@ -105,6 +105,56 @@ RELATE_EDITABLE_INST_ID_BEFORE_VERIFICATION = True
 
 # }}}
 
+
+# {{{ user full_name format
+
+# RELATE's default fullname format is "'%s %s' % (first_name, last_name)",
```
inducer: Looks like this has an extra comma.

local_settings.py.example
```diff
@@ -105,6 +105,56 @@ RELATE_EDITABLE_INST_ID_BEFORE_VERIFICATION = True
...
+
+# }}}
+
+# {{{ system email appelation priority
+
+# RELATE's default email appelation of the receiver is a ordered list:
+# ["first_name", "email", "username"], when first_name is not None
+# (e.g, first_name = "Foo"), the email will be opened
+# by "Dear Foo,". If first_name is None, then email will be used
+# as appelation, so on and so force.
```
inducer: force -> forth

local_settings.py.example
```diff
@@ -105,6 +105,56 @@ RELATE_EDITABLE_INST_ID_BEFORE_VERIFICATION = True
...
+# }}}
+
+# {{{ 
+
+# When displayed in lists, user realname is diplayed by two fields,
+# last_name and first_name, so that the bpth fields can be ordered.
+# You can set the following configuration to True if you want to
+# display name using full_name format.
+
+#RELATE_DISPLAY_USER_FULLNAME_IN_LIST = True
```
inducer: Can you explain this?

inducer:
I like this patch. I think it's the best approach to name formatting we've discussed so far.

course/admin.py
dzhuang: sorry, that.modification should.have been reverted.

local_settings.py.example
dzhuang: I add this configuration because gradebook-participants, e.g., [this page](https://relate.cs.illinois.edu/course/CS210/grading/participants/), which is 404 to me, are displaying user names using last_name and first_name, I don't know if you intended to display in this way so that you can sort user by their last_name or first_name. But for some user like me, I just want to sort user by their full_name since it can be searched. So I add ``RELATE_DISPLAY_USER_FULLNAME_IN_LIST``, so that it display full_name instead of two separated fields. I think you are confused because of the naming.

inducer: OIC. Honestly, just remove that setting. Search is good enough.

local_settings.py.example
```diff
@@ -105,6 +105,56 @@ RELATE_EDITABLE_INST_ID_BEFORE_VERIFICATION = True
...
+# username.
+
+#RELATE_EMAIL_APPELATION_PRIORITY_LIST = ["first_name", "email", "username", "full_name"]
+
+# }}}
+
+# {{{ 
+
+# When displayed in lists, user realname is diplayed by two fields,
+# last_name and first_name, so that the bpth fields can be ordered.
```
inducer: bpth -> both

local_settings.py.example
dzhuang: I just copied it from [here](https://github.com/inducer/relate/blob/master/accounts/models.py#L142)

dzhuang:
I've test it and it should ok now.

inducer:
Looks great! Thanks for submitting this, and for incorporating my feedback!

PR #156, opened 2016-01-29
Add attributes role

lukeolson:
@inducer, this adds the ability to include roles in the `.attributes.yml`:
```
public:
    - test1.pdf
student:
    - test2.pdf
```

The roles are hierarchical in that `instructor` would still be able to view both `test1.pdf` and `test2.pdf` above.  And `public` cannot view `test2.pdf`.


lukeolson:
I'm closing to rebase with uiuc

PR #158, opened 2016-01-30
fix TEMPLATE_* warnings

lukeolson:
this was a simple removal of TEMPLATE_*.  Mostly removing redundancy
that was causing the warnings.


inducer:
lgtm, thx!

PR #159, opened 2016-01-30
local select2 'static' resources instead of from CDN

dzhuang:
https://github.com/applegrew/django-select2/blob/89674ec375291277c5fd17b57be3f4c74b66b83a/django_select2/conf.py


inducer:
Awesome, thank you!

PR #164, opened 2016-02-03
i18n of select2

dzhuang:
The i18n files of most js packages (not for django) are name using format ll-CC. ll for locale (lower case), and CC for country (Uppercase).  and django's ``get_current_language`` templatetags get a language name ll-cc.  This won't cause problem on windows machine, when loading a static i18n file from a js package, since windows console are not case sensitive. However, for linux server, it will result in 404 because it's case sensitive.

I borrowed from built-in class and created a new templatetag ``get_current_js_lang_name`` to fix the problem.

### DIFF
--- a/course/auth.py
+++ b/course/auth.py
@@ -175,10 +175,9 @@ def impersonate(request):
     else:
         form = ImpersonateForm(request.user, language=language)
 
-    return render(request, "generic-form.html", {
+    return render(request, "select2-form.html", {
         "form_description": _("Impersonate user"),
-        "form": form,
-        "select2_i18n": True
+        "form": form
         })
 
 
--- a/course/auth.py
+++ b/course/auth.py
@@ -124,6 +124,7 @@ class ImpersonateMiddleware(object):
 
 class ImpersonateForm(StyledForm):
     def __init__(self, impersonator, *args, **kwargs):
+        language = kwargs.pop("language", None)
         super(ImpersonateForm, self).__init__(*args, **kwargs)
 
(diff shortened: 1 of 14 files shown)


dzhuang:
In fact I should call it l10n of select2?

inducer:
> For the 3rd commit, should I call it l10n of select2?

Nah, that's OK. The current message is close enough to be able to see what's going on.

inducer:
This looks good in principle. Yet, [this](https://github.com/select2/select2/blob/master/src/js/select2/i18n/en.js) makes it look like the only localized bit that a RELATE user will ever see is "no results found". (Since we don't use Select2 in AJAX mode) So it makes me wonder about the cost of maintaining this vs. the benefit. Don't get me wrong--I'm not rejecting this, I just want to accurately understand what we're gaining by doing this, even if the answer ends up being "not much".

dzhuang:
Yes, you're right, at least for now. What I am doing is to enable i18n or l10n of js packages (if those functions are provided) in RELATE. Datatable is an example, and select2 is another. But the latter must combine tweak in python code, that makes it expensive. I think may be in the future we will integrate more packages which are not using django styled i18n files.

dzhuang:
What's more. I think it's bad to modify the base.html. I'll try to clean the code for the 3rd commit.

dzhuang:
I think carefully about your advice. I am indeed using too much resources in enable the i18n of select2, and if in future the dependent package is upgraded, it will brought trouble for us to maintain. I will close this PR while provide a reverted version which still enable the templatetag ``get_current_js_lang_name``.

PR #166, opened 2016-02-04
Templatetag get_current_js_lang_name and i18n of DataTables.

dzhuang:
The i18n files of most js packages (not for django) are name using format ll-CC. ll for locale (lower case), and CC for country (Uppercase).  and django's ``get_current_language`` templatetags get a language name ll-cc.  This won't cause problem on windows machine, when loading a static i18n file from a js package, since windows console are not case sensitive. However, for linux server, it will result in 404 because it's case sensitive.

I borrowed from built-in class and created a new templatetag ``get_current_js_lang_name`` to fix the problem. 

What's more, the tag is applied to DataTables in this PR.

### DIFF
--- a/course/templates/course/gradebook-by-opp.html
+++ b/course/templates/course/gradebook-by-opp.html
@@ -153,13 +153,14 @@
     </tbody>
   </table>
 
-  {% get_current_language as LANGUAGE_CODE %}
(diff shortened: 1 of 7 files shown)


inducer:
Looks great, thanks!

PR #169, opened 2016-02-04
l10n of datetime. fix #167

### LINKED ISSUE #167: Datetime shown not localized.
As one of my student complaints, his browser is using en_US locale, however, the datetime shown in RELATE is in zh_CN format. Although he can understand Chinese, however, that will cause problem in other circumstances. 

The reason of problem lies in [this PR](https://github.com/inducer/relate/pull/39) I provided. It uses setting.LANGUAGE_CODE to determine the locale show in browser. 

I found I've made it too complicated in that PR, and it quite simple to modify that method using built-in method ``django.utils.formats``, and the result is itself locale aware.

I'll create a PR later to fix it.

dzhuang:


inducer:
Looks good. Two comments:

* I'd like `RELATE_CUSTOM_DATETIME_FORMAT_METHOD` mentioned somewhere in `local_settings.py.example`, since I don't think having undocumented "submarine" options floating around is that good.
* Doesn't this remove our dependency on `babel`?

dzhuang:
I think it's dirty to use a custom formatting method. It will result in different rendered format between python and template. The best way is to use templatetags or django format. So I removed the custom formatting method.

dzhuang:
Updated according to [built-in date filter](https://github.com/django/django/blob/be9bd3348d3c4b060b6eea53ef23d0651c122c77/django/template/defaultfilters.py#L725)

inducer:
Looks great. Thank you!

PR #176, opened 2016-02-13
update mime-type to include text/plain

lukeolson:
This adds `text/plain` to the list of mime-types so that it's viewable in the browser.  `octet-stream` does not allow viewing.

I also cleaned up a bunch of spacing.  This makes the PR difficult to read.  It's basically a one-line change + the docstring.


inducer:
Merged in `520d96e`.

inducer:
(Also now live on the site.)

PR #177, opened 2016-02-24
Change the grading buttons so that the scale is out of the right number of points.

mattwala:


inducer:
Thanks for looking into this. Unfortunately, this isn't quite right as it is.  `human_feedback_point_value` isn't part of the `Page` interface:

https://github.com/inducer/relate/blob/master/course/page/base.py#L201

This all (including the clicky Javascript stuff that's already there) goes a bit against the grain of the page interface, because in principle a page is free to show whatever Django form it wants for grading--that's what `make_grading_form` et al are for. Ideally, the JS would be generated by this base class:

https://github.com/inducer/relate/blob/master/course/page/base.py#L687

using this machinery:

https://docs.djangoproject.com/en/1.9/topics/forms/media/

(and this subclass then also knows about `human_feedback_point_value`)

mattwala:
@inducer, is this closer to what you mean? I am not quite happy with the way that it hard-codes some of the button classes when it generates the widget, but I am not sure how to separate that out.

inducer:
Looks great! Thanks for doing this.

inducer:
(Change is also live.)

PR #180, opened 2016-02-28
Sandbox: Store the page_data in the session along with the answer_data.

mattwala:
Otherwise, the answer_data could be inconsistent with the page_data.
For instance, when sandboxing a choice question that is shuffled, the
answer_data is matched with a particular permutation which is part of
the page_data.


mattwala:
I just noticed this has a bug, unfortunately. Revising.

mattwala:
Fixed the bug.

inducer:
Looks good. Thanks for the contribution!

PR #183, opened 2016-03-04
Raise form ValidationError when importing non-importable csv data

dzhuang:
I failed to find a satisfying solution that can smartly detect the codec of csv files which are converted from MS Excel. The current solution is to raise form ValidationError, when user are importing csv with columns to be imported containing non-ASCII characters while the csv is not utf-8 encoded.


inducer:
Looks great. Thanks for the contribution!

PR #184, opened 2016-03-05
Use JavaScript localeCompare to sort "Name" field in dataTable.

dzhuang:
It helps in sorting user full names which are not using English characters (e.g., Chinese), and some languages which won't provide correct ordering with alphabetical sorting (e.g., Spanish).

### DIFF
--- a/course/templates/course/gradebook.html
+++ b/course/templates/course/gradebook.html
@@ -50,7 +50,8 @@
         "scrollX": true,
         "scrollCollapse": true,
         "paging": true,
-        "ordering": false,
+        "ordering": true,
+        "columnDefs": [{ type: 'name', targets: 1 }],
         "language": {url: '{% static "datatables-i18n/i18n/" %}{{LANG}}.json'},
     } );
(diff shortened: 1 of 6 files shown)


inducer:
`participation-table.html` gets used from the "participation query" function, too--does this break that?

dzhuang:
I think not. That seems to only change how the tables are rendered, without affecting the query results.

course/templates/course/gradebook.html
```diff
@@ -50,7 +50,8 @@
         "scrollX": true,
         "scrollCollapse": true,
         "paging": true,
-        "ordering": false,
```
dzhuang: I have no idea why ``"ordering"`` is originally set to be ``false``. It works well if set to be ``true``.

inducer:
Looks good. Thanks!

dzhuang:
I think I figure out why you originally disable sorting in the gradebook-- the sorting of grade opportunity won't give expected result?

PR #185, opened 2016-03-05
Display username in ImpersonateForm.

dzhuang:
Currently, username is not displayed (and consequently not searchable) when impersonating users. That is quite inconvenient when debugging in some circumstances.


inducer:
Looks good. Thanks!

Issue #186, opened 2016-03-08
Sandbox Errors


dzhuang:
It raises ScannerError: mapping values are not allowed here in "", line 2, column 11: id: eigvec ^

The testing code is from [sample](https://github.com/inducer/relate-sample/blob/master/flows/quiz-test.yml)
```
        type: TextQuestion
        id: eigvec
        value: 2
        prompt: |
            # What's an eigenvector?
            Yadda ___________________ yadda.
        answers:

          - <plain>matrix
          - <case_sens_plain>Eigenmatrix
          - <regex>(?:linear\s+)?\s*map
          - <case_sens_regex>(?:operator\s+)?\s*map

```

inducer:
You need to remove the indentation.

dzhuang:
I remembered originally it won't require to remove indentation. I think the previous way is better than what it is now. In the past, I just copied working snippet directly from sandbox to my .yml, without more editing need.

inducer:
Can you cook up a patch?

Here's a piece that might be useful:
https://github.com/inducer/pytools/blob/master/pytools/py_codegen.py#L98

dzhuang:
Thank you for the great suggestion!

dzhuang:
Should be fixed by #187

PR #187, opened 2016-03-09
Remove common indentation of contents before yaml.load in page sandbox.

dzhuang:


course/sandbox.py
```diff
@@ -344,4 +345,35 @@ def make_form(data=None):
...
+def remove_sb_content_common_indentation(code):
+    if "\n" not in code:
+        return code
+
+    lines = code.split("\n")
+
+    if lines:
+        base_indent = 0
+
+        # form field with textarea widget returns
```
dzhuang: Maybe I am wrong in that the first line is stripped by codemirror.CodeMirrorTextarea?

course/sandbox.py
```diff
@@ -344,4 +345,35 @@ def make_form(data=None):
...
+
+    lines = code.split("\n")
+
+    if lines:
+        base_indent = 0
+
+        # form field with textarea widget returns 
+        # cleaned_data with line[0] left stipped,
+        # so here we are starting from line[1].
+        while lines[1][base_indent] in " \t":
```
dzhuang: It will fail when line[1] is " "?

dzhuang:
The snippet is not OK, please don't merge.

dzhuang:
The reason for #186 is my recent upgrade of Django from 1.8.x to 1.9.2, in which ``CharField`` add a new attribute ``strip`` with default value ``True`` [(link)]    (https://github.com/django/django/commit/11cac1bd8ef7546ca32d9969d4348bf412dc6664#diff-2645e84b6e5bc956ae62fe5f0a7d60faR372) .

BTW, it is surprising "whether or not strip by default for CharField content" had been discussed for almost 9 years!

inducer:
Looks good. Thanks for tracking this down!

PR #189, opened 2016-03-14
Allow import grade by csv using participant's institutional id.

dzhuang:
Along with more exception handling when importing csv.

### DIFF
--- a/course/grades.py
+++ b/course/grades.py
@@ -1110,11 +1110,11 @@ def csv_to_grade_changes(
 
                 updated = []
                 if last_grade.points != gchange.points:
-                    updated.append(_("points"))
+                    updated.append(ugettext("points"))
                 if last_grade.max_points != gchange.max_points:
-                    updated.append(_("max_points"))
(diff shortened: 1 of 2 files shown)


course/grades.py
```diff
@@ -1033,8 +1065,12 @@ def csv_to_grade_changes(
         gchange = GradeChange()
         gchange.opportunity = grading_opportunity
         try:
-            gchange.participation = find_participant_from_id(
-                    course, row[id_column-1])
+            if attr_type == "email_or_id":
+                gchange.participation = find_participant_from_id(
+                        course, row[attr_column-1])
+            else:
```
inducer: Can you change this logic to be `elif attr_type == "...." else: raise`?

inducer:
Looks good aside from the minor wrinkle I pointed out.

dzhuang:
Updated. Thanks for the suggestions.

inducer:
Looks good. Thanks!

PR #192, opened 2016-03-15
allow faking time when impersonating.

dzhuang:


inducer:
Please also check whether the original user is staff, although that may appear redundant at this point.

dzhuang:
Is this patch ok or any improvement can be made?

inducer:
Close enough. Personally, I'd get the "actual" user and check their permissions once rather than duplicating that code. But this is OK.

PR #194, opened 2016-03-17
no save button for file upload question.

dzhuang:
Along with mime-type filter for browser when selecting file.


inducer:
It's not pretty, but it's also not overly hurtful--so why not.

inducer:
Thanks.

dzhuang:
Sorry for that ugly fix :-), and I think your suggestions helped me a lot, thanks!

PR #197, opened 2016-04-16
Gradebook tweaks

mattwala:
Makes the following changes to the per-opportunity gradebook:
* Sorts by user id by default
* Shows all flow sessions
* Adds a "View page grades" button which allows all page grades to be shown in the gradebook

### DIFF
--- a/course/templates/course/gradebook-by-opp.html
+++ b/course/templates/course/gradebook-by-opp.html
@@ -104,13 +104,16 @@
       <th class="datacol">{% trans "Overall grade" %}</th>
       <th class="datacol">{% trans "Session state" %}</th>
       <th class="datacol">{% trans "Session grade" %}</th>
+      {% for i in page_numbers %}
+        <th class="datacol">{{ i }}</th>
+      {% endfor %}
     </thead>
     <tbody>
       {% for participation, grade_info in grade_table %}
       <tr>
-        <td class="datacol">
+        <th class="headcol">
           <a href="{% url "relate-view_single_grade" course.identifier participation.id opportunity.id %}"><span class="sensitive">{{ participation.user.username }}</span></a>
-        </td>
+        </th>
         <td class="datacol">
           <span class="sensitive">
             {{ participation.user.last_name }}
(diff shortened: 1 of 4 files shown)


mattwala:
* Also re-orders the columns so that the overall grade is the last column

inducer:
Looks great! Thanks for the patch. This is now live on the UIUC instance.

One question: Why did you decide to show the page grade at the end of the page columns rather than at the beginning? Any particular reason or coincidence?

mattwala:
I'm not sure what you mean. What page columns are you referring to?

inducer:
The columns with the per-page grades precede the overall grade. Instinctively, I would have ordered them the other way around.

mattwala:
Ah, I suppose if there is any reason why I chose it that way, it is because the "Overall grade" column is not that useful from the grader's perspective. As a grader I would want to have the page grades in the center and the "Overall grade" just uses up real estate and can be confused for a page grade.

mattwala:
I suppose putting it as the first column would also be fine.

inducer:
OIC. Fine as it is in this case. Thanks again!

course/templates/course/gradebook-by-opp.html
```diff
@@ -109,12 +111,17 @@ <h3 class="start-well-title">{% trans "Modify many sessions at once" %}</h3>
           <a href="{% url "relate-view_single_grade" course.identifier participation.id opportunity.id %}"><span class="sensitive">{{ participation.user.username }}</span></a>
         </td>
         <td class="datacol">
-            <span class="sensitive">
-                {{ participation.user.get_full_name }}
```
dzhuang: I am wondering why the user.get_full_name here is reverted back to 2 columns ;-(

inducer: @dzhuang Sorry, missed that. @mattwala Could you send a PR reverting that?

mattwala: Done. #198

inducer: Merged.

PR #201, opened 2016-04-25
fix #200.

### LINKED ISSUE #200: Automatically redirect back to previously visited page after login
In cases when users are required to login/register.

dzhuang:
fix #200 

Currently, only sign in by pw is tweaked.


inducer:
Looks great. Thanks!

PR #202, opened 2016-04-26
add user and impersonator to FlowPageVisit.

### LINKED ISSUE #199: FlowPageVisit bug?
After the instructor/TAs visited an attendant's flow pages, ``FlowPageVisit`` instances will be saved as if it is that attendant who had made those visits. I don't know if it is actually a bug or it is intended to be designed in this way. 

If it's a bug, then I think it's better that when impersonating, there should be no new instance saved or save the user who is impersonating as the ``participant``.

*Update*:
I am wrong in that actually ``FlowPageVisit`` doesn't have a property recording the visitor, the ``participant`` property displayed in Admin is from the related ``flow_session.participation.user``. I think it's better not to save those instances when the visit is make by course staff.

For impersonating, however, if the course staff submit answer for the attendants, is it possible to distinguish those visits from the actual visits made by the participants themselves?

dzhuang:
Fix #199


inducer:
Looks great! Thanks for the contribution.

PR #207, opened 2016-05-05
add a latex macro to convert Tex code to <img> tag using dataURI.

dzhuang:
Requirement:
 - latex
 - dvipng (included in Textlive)
 - dvisvgm

USAGE:

```
{% call latex() %}
The latex code
{% endcall %}
```

Optional args:

- ``output_dir``: <em>full path</em> or RELATE styled path of the file generated.
- ``tex_filename``: the based filename of the latex,  and image as well, if not set, use md5 of the full latex code
- ``image_format``: the output format of the image, only png and svg are available, and svg is the default. If set png, code with tikz/pgf will still using svg.
- ``tex_preamble``: the preamble other than what is provided in settings or default value in course.latex_utils. In the yml, it can be set by using {% set foo %}{% endset %} block. 
- ``tex_preamble_extra``: more packages or settings appended to default preamble.
- ``overwrite``: regenerate image if it exists, default: False(recommended).
- ``is_full_code``: whether the latex code is full code (containing preamble), default: False
- ``html_class_extra``: extra html class for the  tag, besides ``img-responsive``
- ``alt``: alt attribute for the  tag. default value is the code in the call block.

### DIFF
--- a/course/content.py
+++ b/course/content.py
@@ -759,6 +759,14 @@ def markup_to_html(course, repo, commit_sha, text, reverse_func=None,
     env = Environment(
             loader=GitTemplateLoader(repo, commit_sha),
             undefined=StrictUndefined)
+
+    from course.latex_utils import tex2imgtag
+
+    def jinja_tex2imgtag(caller, *args, **kwargs):
+        return tex2imgtag(caller(), *args, **kwargs)
+
+    env.globals["latex"] = jinja_tex2imgtag
+
     template = env.from_string(text)
     text = template.render(**jinja_env)
 
--- a/course/latex_utils.py
+++ b/course/latex_utils.py
@@ -1,0 +1,400 @@
+# -*- coding: utf-8 -*-
+
+from __future__ import division
+
+__copyright__ = "Copyright (C) 2016 Dong Zhuang, Andreas Kloeckner"
+
+__license__ = """
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in
+all copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+THE SOFTWARE.
+"""
+
+import six
+import os
+import re
+import errno
+from subprocess import Popen, PIPE
+
+from django.utils.translation import ugettext as _
+from django.conf import settings
+
+# {{{ latex 2 image conversion
+
+ALLOWED_LATEX2IMG_FORMAT = ['png', 'svg']
+
+DEFAULT_LATEX_PREAMBLE = r'''
+\documentclass{article}
+\usepackage[fontsize=14px]{scrextend}
+\usepackage{amsmath}
+\usepackage{amsthm}
+\usepackage{amssymb}
+\usepackage{tikz}
+\usepackage{pgf}
+\usepackage{CJKutf8}
+\usepackage{bm}
+\pagestyle{empty}
+'''
+
+DEFAULT_LATEX_IMAGE_FOLDER_NAME = getattr(
+    settings, "RELATE_LATEX_IMAGE_FOLDER_NAME", "latex_image")
+
+LATEX_LOG_OMIT_LINE_STARTS = (
+    "See the LaTeX manual or LaTeX",
+    "Type  H <return>  for",
+    " ...",
+    # more
+)
+
+LATEX_ERR_LOG_BEGIN_LINE_STARTS = "\n! "
+LATEX_ERR_LOG_END_LINE_STARTS = "\nHere is how much of TeX's memory"
+
+PAGE_EMPTY_RE = re.compile(r"\n[^%]*\\pagestyle\{empty\}")
+TIKZ_PGF_RE = re.compile(r"\n[^%]*\\begin\{(tikzpicture|pgfpicture)\}")
+
+
+# {{{ file read and write
+
+def _file_read(filename):
+    '''Read the content of a file and close it properly.'''
+    f = file(filename, 'rb')
+    content = f.read()
+    f.close()
+    return content
+
+
+def _file_write(filename, content):
+    '''Write into a file and close it properly.'''
+    f = file(filename, 'wb')
+    f.write(content)
+    f.close()
+
+# }}}
+
+
+# {{{ assemble tex source
+
+def make_tex_source(tex_body, tex_preamble="",
+                    tex_preamble_extra="", is_full_code=False):
+    '''
+        Assemble tex source code.
+        If is_full_code is True, tex_body it self contains
+        preamble, which make it convenient for pgf/tikz
+        settings.
+    '''
+    assert isinstance(tex_body, unicode)
+
+    # In case tex_body contains preamble
+    if is_full_code:
+        tex_source = tex_body
+    else:
+        if not tex_preamble:
+            tex_preamble = getattr(
+                settings, "RELATE_LATEX_PREAMBLE",
+                DEFAULT_LATEX_PREAMBLE)
+
+        # allow wrap tex_body by some environments
+        tex_begin_document =getattr(
+            settings, "RELATE_LATEX_BEGIN_DOCUMENT",
+            r"\begin{document}")
+
+        tex_end_document =getattr(
+            settings, "RELATE_LATEX_END_DOCUMENT",
+            r"\end{document}")
+
+        tex_source = "%s" * 5 % (
+            tex_preamble, tex_preamble_extra,
+            tex_begin_document, tex_body, tex_end_document)
+
+    # make sure the latex source use empty style (no page mark)
+    if not re.search(PAGE_EMPTY_RE, tex_source):
+        tex_source = tex_source.replace(
+            "\\begin{document}",
+            "\n\\pagestyle{empty}\n\\begin{document}")
+
+    return tex_source
+
+# }}}
+
+
+def get_abstract_latex_log(log):
+    '''abstract error msg from latex stdout'''
+
+    msg = log.split(LATEX_ERR_LOG_BEGIN_LINE_STARTS)[1]\
+        .split(LATEX_ERR_LOG_END_LINE_STARTS)[0]
+
+    if LATEX_LOG_OMIT_LINE_STARTS:
+        msg = "\n".join(line for line in msg.splitlines()
+                        if (not line.startswith(LATEX_LOG_OMIT_LINE_STARTS)
+                            and
+                            line.strip() != ""))
+    return msg
+
+
+# {{{ covert tex to dvi format
+
+def tex2dvi(tex_source, output_dir, basename, overwrite=False,
+            max_runs=5):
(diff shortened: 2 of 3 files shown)


dzhuang:
Please don't merge, I think the is_full_source arg can be removed. it should be more clever in checking whether the latex code is full or not.

dzhuang:
Updated, and looking forward for your advices.

inducer:
Some comments from a first read:

* I'm not sure I like that the preamble is configurable per-site. That means course content would become site-specific, because different sites may use different preambles.
* There's a security implication here. We're running a bunch of commands with potentially semi-trusted input. At least `latex` should be run with `-no-shell-escape`. What about `\verbatiminput{/etc/passwd}`?
* `tex2...I` - PEP8
* Why make this DVI-based? That seems backwards. Why not PDF?
*  Raising an exception because something is wrong with user-written content should not happen outside of validation. (Otherwise, say, the course page may not render any more, and it becomes hard to even update the content.)
* I'm not sure I like the idea that this leaves permanent traces on the file system. Why not run in a temporary directory and blow that away when done. Also, as is you assume you know about every file produced by TeX, which is generally not true. If you'd like to cache, use django's cache mechanism.

dzhuang:
About site-specific preamble. Actually, I had realized that problem when writing the code, as you can see I tried to make preamble page-specific in the code, and I need your advice on where (and probably how) to elegantly configure the preamble in git so that it can be course-specific.

About DVI-based, personally I prefer pdflatex over latex, but I not sure whether there are some tools which can convert pdf to image that are more convenient than dvi2png and dvi2svgm, which are included in Texlive-full, without the need to consider how to install across platform. Maybe imagemagick? 

Thank you for all the precious advices. I'll try to improve the snippet.

dzhuang:
In fact, my latex compilation setup is somewhat huge, include many packages, custom commands, something like makeatletter, tikz settings, and many external image files. My idea follows:
 1. Configure all the needed path like preamble.tex, package.tex, config.tex and figure directory in the course git, like attribute.xml
 1. Make git root of the course the default working directory, copy those files to WD if they don't exist or don't match.
 1. Use temp directory in wd when compiling, and blow it away when done as you advised.

In this way, I can merge my local latex project into the course, without keeping 2 different versions.

P.S., I've been trying these days to find a *crossplatform* pdf2png/pdf2svg convertor, but failed, and I found this [post](https://kate-editor.org/2013/06/01/how-to-convert-pdf-to-svg/) also complainted about this issue and stay using dvi2png and dvi2svg as the solution. Crossplatform is import for me, because I am developing on windows while deploy on linux.

dzhuang:
Now I think it's better for the latex code in a page independent of external resources, because if those external settings changed, the previously generated images might fail. The best practice is to offer the full document in the page (at least all elements can be assembled in that page).

dzhuang:
I now accept that we should use pdf as base to generate the image, as least as an option. From my experience these days, I found it painful in configuring the fonts and character set on the linux server, just for the purpose of displaying Chinese characters. 

Svg is good format, and use pdf to generate svg is hard, at least for now.  But using dvi as base will  bring problems in the future. For example, if I want to migrate data from one server to another, I'll have to do the complex configuration again, that will also be a high barrier for new user.

My current solution is:
  1. Add pdflatex and xelatex compilers, and leaving latex as an option. As xelatex is easier to configure in terms of fonts and character set.
  1. Because use a different compiler to compile the same code might result in failure, user must specify which compiler is used, and there should be no default compiler either at site level or course level. And that also leave room for more compiler to be used in the future.
  1. Use ``imagemagick`` to convert PDFs to png, although the generated images are not scalable and not looking good, but it's really a life saver. Moreover, the option to generate svg from dvi is preserved, leaving possibilities to the generate svgs.
  1. Another thing, don't use alt to display source code, because Mathjax think \begin{tikzpicture} \end{tikzpicture} as math expressions, and will make it a mess while not display the valid images.

@inducer  I need your advice on this, because I am in urgent need to use this feature. As I've already use this feature on my instance, and the pages are growing. I think I need to rewrite some of the code, but before the number page is out of control. For now, I want to determine args/kwargs (required/optional) that I may use in the caller. My current idea is:

Required args:
- ``compiler``: the command line used to compile the tex file, currently available: xelatex, pdflatex and latex.

Optional:
- ``output_dir``: <em>full path</em> or RELATE styled path of the file generated, default a directory named "latex_image" under MEDIA_ROOT.
- ``tex_filename``: the based filename of the latex,  and image as well, if not set, use md5 of the full latex code
- ``image_format``: the output format of the image, only png and svg are available, and png is the default. If ``compiler="latex"``, and if the code contains ``tikz/pgf`` , the image_format will be using svg.
- ``tex_preamble``: the preamble other than what is provided in settings or default value in course.latex_utils. In the yml, it's value might be better to be set by using {% set foo %}{% endset %} block and the use ``tex_premable=foo`` as argument (same for ``tex_preamble_extra``). 
- ``tex_preamble_extra``: more packages or settings appended to default preamble.
- ``overwrite``: regenerate image if it exists, default: False(recommended).
- ``html_class_extra``: extra html class for the  tag, besides ``img-responsive``
- ``alt``: brief description of the image, default: ``tex_filename``.

Currently, I have only used ``image_format`` kwarg, and I want to add the ``compiler`` as the first arg. All the other name of kwargs can be changed, arg.  If no more args are needed, I can set out to alter my code now.  Looking forward for you reply, thanks.

dzhuang:
UPDATE:

 1. Allow pdflatex/xelatex/latex compilation
 1. Use latexmk to determine the number of times needed for tex compilation.
 1. Use Imagemagick to convert PDFs to png.
 1. Use django check framework to check if the path/version of compilers and converters are correctly configured.

Requirement:
   - Texlive-full (TUG version Texlive is prefered since it contains all latest packages, noticing that requires ``RELATE_LATEX_BIN_DIR`` configuration in settings for linux server.)
   - ImageMagick (for windows, that requires ``RELATE_IMAGEMAGIC_BIN_DIR`` configuration in settings.)

USAGE:

```
{% call latex(compiler="pdflatex", image_format="png") %}
The latex code
{% endcall %}
```

Required args:
- ``compiler``: string, the command line used to compile the tex file, currently available: xelatex, pdflatex and latex.
- ``image_format``: string, the output format of the image, only png and svg are available for now. If ``compiler="latex"``, and if the code contains ``tikz/pgf`` , the image_format will be forced using svg.

Optional:
- ``tex_filename``: string, the based filename of the latex,  and image as well, if not set, use md5 of the full latex code
- ``tex_preamble``: string, allow user to split the preamble from the full document. In the yml, it's value might be better to be set by using {% set foo %}{% endset %} block and the use ``tex_premable=foo`` as argument. That parameter can be saved as a macro in the course git, but it's not recommend, as changing the preamble will force the originally valid image to recompile, which might result in failure (same for ``tex_preamble_extra``). The best practice is to keep full latex document, which might be changed in the future, in the page markup.
- ``tex_preamble_extra``: string, more packages or settings appended to ``tex_premable``.
- ``force_regenerate``: boolean, if True, regenerate image if it exists, default: False(recommended).
- ``html_class_extra``: string, extra html class for the  tag, besides ``img-responsive``
- ``alt``: string, brief description of the image, default: the document part of the tex source.

Example is also updated in https://github.com/inducer/relate-sample/pull/4

dzhuang:
UPDATE:

``{{``, ``}}``, ``{%``, ``%}``, ``{#`` and ``#}`` are used in jinja  template, if the calling latex code contains those string, jinja will fail to render. The work around is to manually insert a space (spaces or tabs) between the two character (e.g., ``{{`` --> ``{  {``) for each of those strings appeared in latex code, and then before compiling the latex source, remove the spaces (tabs) added.

course/check.py
```diff
@@ -0,0 +1,53 @@
...
+from django.conf import settings
+
+from course.latex.utils import get_all_indirect_subclasses
+from course.latex.converter import CommandBase
+
+class Tags(DjangoTags):
+    relate_course_tag = 'relate_course_tag'
+
+@register(Tags.relate_course_tag)
+@register(Tags.relate_course_tag, deploy=True)
```
inducer: Does this need to be duplicated?

inducer: I'm not saying it's wrong, in fact I don't know--the duplication just looks fishy.

dzhuang: You are right. I'll correct it.

course/latex/latex.py
```diff
@@ -0,0 +1,176 @@
...
+        :param text: string. Full latex document, or body only if
+        preamble or preamble_extra are given.
+        :param preamble: string. If full document is provided in
+        text, this value will be neglected.
+        :param preamble_extra: string. Append to existing preamle.
+        :param empty_pagestyle: bool. If True, the pagestyle will
+        be set as "empty". We are not using
+        \documentclass{standalone}.
+        """
+        if text:
```
inducer: Could you change this to `if not text: raise`? This avoids a bunch of nesting.

course/latex/latex.py
```diff
@@ -0,0 +1,176 @@
...
+    has_begindoc = False
+    has_enddoc = False
+
+    def is_empty_pagestyle_already(self):
+        match = re.search(r"\\pagestyle{\s?empty\s?}", self.preamble)
+        if match:
+            return True
+        return False
+
+    def parse(self, latex, test=False):
```
inducer: Why is this parsing step needed? It seems fairly brittle.

course/latex/converter.py
```diff
@@ -0,0 +1,646 @@
...
+
+class Imageconverter(CommandBase):
+
+    @property
+    def output_format(self):
+        raise NotImplementedError
+
+    def __init__(self):
+        bin_path_dir = getattr(
+            settings, "RELATE_%s_BIN_DIR" % self.name.upper(),
```
inducer: Should these settings be documented in `local_settings.py.example`?

inducer:
* This needs some documentation (along the lines of what you included above) to be in the `.rst` file.
* How does this manage temporary files? I couldn't find that in the code, but it's very important to do that in a secure and race-free manner.

course/latex/latex.py
dzhuang: OK. I also found line 126-130 should be removed.

course/latex/latex.py
dzhuang: With this, we can call the macro like [this](https://github.com/inducer/relate-sample/pull/4/files#diff-0bdbb3e999a8bd689c5cff408129ecb8R53). If not, ``parse`` will fail because there are no preamble in the call block. Because a full latex doc should contain ``\documentclass``, ``\begin{document}``, ``\end{document}``.

``parse`` using ``test=True`` (which might need to be renamed so as not to cause misunderstanding) is to check whether preamble are defined out side the block, and assemble them into full latex code, and try parse again.

course/latex/converter.py
dzhuang: Yes, it's [here](https://github.com/inducer/relate/pull/207/files/6d2d8f5032874af4f1a26017e3c175ce413f45f8#diff-fa1395737ba4b4fa8329b63aff5be72dR154). Or, did you mean besides ``ImageMagick``, path to ``dvisvgm`` and ``dvipng`` should also be configured?

course/latex/latex.py
```diff
@@ -0,0 +1,176 @@
...
+        \documentclass{standalone}.
+        """
+        if text:
+            text = strip_comments(text)
+            try:
+                self.parse(text)
+            except TexDocMissingElementError:
+                self.parse(text, test=True)
+                if self.has_preamble:
+                    # begin_document or end_document is missing
```
dzhuang: Because ``TexDocMissingElementError`` will be raise if only one of ``\documentclass``, ``\begin{document}`` and ``\end{document}`` is missing.

dzhuang:
For temporary files, I think they are created [here](https://github.com/inducer/relate/pull/207/files#diff-57ac0b94f7c1b27bced27689bd3c3a0fR388), and removed using ``_remove_working_dir`` defined [here](https://github.com/inducer/relate/pull/207/files#diff-57ac0b94f7c1b27bced27689bd3c3a0fR378).

dzhuang:
It has been 2 month since last comment, how time flies. Sorry, I'll return to this asap.

dzhuang:
Update:
- Add documentation
- Enable lualatex

course/latex/converter.py
```diff
@@ -0,0 +1,688 @@
...
+            try:
+                log = _file_read(log_path)
+            except OSError:
+                # no log file is generated
+                self._remove_working_dir()
+                raise RuntimeError(error)
+
+            try:
+                log = get_abstract_latex_log(log)
+                _file_write(self.errlog_saving_path, log)
```
dzhuang: Risk of race condition.

course/latex/converter.py
```diff
@@ -0,0 +1,688 @@
...
+        elif n_images > 1:
+            raise ValueError(
+                string_concat(
+                    "%s images are generated while expecting 1, "
+                    "possibly due to long pdf file."
+                    % (n_images, )
+                ))
+
+        try:
+            shutil.copyfile(image_path, self.image_saving_path)
```
dzhuang: Risk of race condition.

Issue #135, opened 2015-11-30
Python code submission fails


duwhop:
Hello I am having trouble submitting python code from question 8 quiz.  I get the following error:

Autograder feedback
The grading code failed. Sorry about that. The staff has been informed, and if this problem is due to an issue with the grading code, it will be fixed as soon as possible. In the meantime, you'll see a traceback below that may help you figure out what went wrong.

This is the exception traceback:

Traceback (most recent call last):
  File "C:\Users\kjeanpie\Desktop\SAILS\relate\course\page\code.py", line 565, in grade
    run_timeout=self.page_desc.timeout)
  File "C:\Users\kjeanpie\Desktop\SAILS\relate\course\page\code.py", line 250, in request_python_run_with_retries
    result = request_python_run(run_req, run_timeout, image=image)
  File "C:\Users\kjeanpie\Desktop\SAILS\relate\course\page\code.py", line 109, in request_python_run
    user="runpy")
  File "C:\python27_x64\lib\site-packages\docker\client.py", line 245, in create_container
    return self.create_container_from_config(config, name)
  File "C:\python27_x64\lib\site-packages\docker\client.py", line 252, in create_container_from_config
    res = self._post_json(u, data=config, params=params)
  File "C:\python27_x64\lib\site-packages\docker\clientbase.py", line 128, in _post_json
    return self._post(url, data=json.dumps(data2), **kwargs)
  File "C:\python27_x64\lib\site-packages\docker\clientbase.py", line 83, in _post
    return self.post(url, **self._set_request_timeout(kwargs))
  File "C:\python27_x64\lib\site-packages\requests\sessions.py", line 508, in post
    return self.request('POST', url, data=data, json=json, **kwargs)
  File "C:\python27_x64\lib\site-packages\requests\sessions.py", line 465, in request
    resp = self.send(prep, **send_kwargs)
  File "C:\python27_x64\lib\site-packages\requests\sessions.py", line 573, in send
    r = adapter.send(request, **kwargs)
  File "C:\python27_x64\lib\site-packages\requests\adapters.py", line 370, in send
    timeout=timeout
  File "C:\python27_x64\lib\site-packages\requests\packages\urllib3\connectionpool.py", line 544, in urlopen
    body=body, headers=headers)
  File "C:\python27_x64\lib\site-packages\requests\packages\urllib3\connectionpool.py", line 349, in _make_request
    conn.request(method, url, **httplib_request_kw)
[...]

PR #212, opened 2016-06-21
Add Docker-machine support for Windows/Mac

dzhuang:
Fixing #135. Tested on Windows, needing tests on Mac.

### DIFF
--- a/course/page/code.py
+++ b/course/page/code.py
@@ -99,7 +99,7 @@ def request_python_run(run_req, run_timeout, image=None):
                 None)
 
         import platform
-        if not platform.system().lower().startswith("win"):
+        if platform.system().lower().startswith("linux"):
             docker_cnx = docker.Client(
                     base_url=docker_url,
                     tls=docker_tls,
--- a/course/page/code.py
+++ b/course/page/code.py
@@ -74,6 +74,7 @@ class InvalidPingResponse(RuntimeError):
 
 
 def request_python_run(run_req, run_timeout, image=None):
+    import platform
     import json
     from six.moves import http_client
     import docker
@@ -98,7 +99,6 @@ def request_python_run(run_req, run_timeout, image=None):
         docker_tls = getattr(settings, "RELATE_DOCKER_TLS_CONFIG",
                 None)
 
-        import platform
         if platform.system().lower().startswith("linux"):
             docker_cnx = docker.Client(
                     base_url=docker_url,
@@ -146,11 +146,11 @@ def request_python_run(run_req, run_timeout, image=None):
                     ["NetworkSettings"]["Ports"]["%d/tcp" % RUNPY_PORT])
             port_host_ip = port_info.get("HostIp")
 
-            if platform.system().lower().startswith("win"):
-                connect_host_ip = getattr(settings, "RELATE_DOCKER_HOST_IP")
-            else:
+            if platform.system().lower().startswith("linux"):
                 if port_host_ip != "0.0.0.0":
                     connect_host_ip = port_host_ip
+            else:
+                connect_host_ip = getattr(settings, "RELATE_DOCKER_HOST_IP")
 
             port = int(port_info["HostPort"])
         else:
--- a/course/page/code.py
+++ b/course/page/code.py
@@ -97,11 +97,20 @@ def request_python_run(run_req, run_timeout, image=None):
                 "unix://var/run/docker.sock")
         docker_tls = getattr(settings, "RELATE_DOCKER_TLS_CONFIG",
                 None)
-        docker_cnx = docker.Client(
-                base_url=docker_url,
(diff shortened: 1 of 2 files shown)


local_settings.py.example
```diff
@@ -174,6 +174,32 @@ RELATE_DOCKER_TLS_CONFIG = None
...
+                raise RuntimeError(
+                    "error %s: %s" % (
+                            error_code, stderr.decode('utf-8').strip()))
+            else:
+                # fail silently
+                return 'localhost'
+        out = stdout.decode('utf-8')
+
+        import re
+        ip_re_pattern = "DOCKER_HOST=tcp://([\d.]+):"
```
inducer: This doesn't cover DNS names or IPv6.

local_settings.py.example
```diff
@@ -174,6 +174,32 @@ RELATE_DOCKER_TLS_CONFIG = None
 #     ca_cert=os.path.join(pki_base_dir, "ca.pem"),
 #     verify=True)
 
+# The following lines are for docker-machine on windows
+import platform
+if not platform.system().lower().startswith("linux"):
```
inducer: I'd rather have this function sitting somewhere in a module within RELATE, to be imported into the config. That way, we can make changes without every user having to update their config. I'd also prefer a naming like `get_docker_host_ip_from_docker_machine` or some such.

course/page/code.py
```diff
@@ -97,11 +98,19 @@ def debug_print(s):
...
+                    base_url=docker_url,
+                    tls=docker_tls,
+                    timeout=docker_timeout,
+                    version="1.19")
+        else:
+            from docker.utils import kwargs_from_env
+            docker_cnx = docker.Client(
                 timeout=docker_timeout,
-                version="1.19")
+                **kwargs_from_env(assert_hostname=False)
```
inducer: Can you explain this?

course/page/code.py
```diff
@@ -137,8 +146,11 @@ def debug_print(s):
                     ["NetworkSettings"]["Ports"]["%d/tcp" % RUNPY_PORT])
             port_host_ip = port_info.get("HostIp")
 
-            if port_host_ip != "0.0.0.0":
-                connect_host_ip = port_host_ip
+            if platform.system().lower().startswith("linux"):
+                if port_host_ip != "0.0.0.0":
+                    connect_host_ip = port_host_ip
+            else:
+                connect_host_ip = getattr(settings, "RELATE_DOCKER_HOST_IP")
```
inducer: *   I'd rather have the variable be named `RELATE_DOCKER_HOST`.
*   I'd like the variable to be used if it's defined. (not just "if it's not LInux"--that seems hard to debug and document if something goes wrong with that setting)

course/page/code.py
dzhuang: First I think I didn't make it clear that this PR is for developing on Windows or Mac OS using a local VM. The documentation of  ``kwargs_from_env()`` can be found [here](https://docker-py.readthedocs.io/en/1.2.3/boot2docker/), the kwarg ``assert_hostname=False`` is used because we are visiting docker from out side the VM, if that kwarg is missing, we will get an error ``CertificateError: hostname '192.168.99.10x' doesn't match 'localhost'``.

I also missed a instruction when using this, for windows, before we run ``python manage.py runserver``, we should run a shell cmd ``@FOR /f "tokens=*" %i IN ('docker-machine env default') DO @%i``. For Mac os, the cmd is ``$eval('docker-machine env default')``, so that ``kwargs_from_env()`` can get correct env variables for running docker.

The env variables on my machine are:

```
SET DOCKER_TLS_VERIFY=1
DOCKER_HOST=tcp://192.168.99.100:2376
DOCKER_CERT_PATH=d:\Users\dzhuang\.docker\machine\machines\default
DOCKER_MACHINE_NAME=default
```

Those variables can be get using cmd ``docker-machine env default`` when ``default`` is the name of the docker-machine.

course/page/code.py
dzhuang: You are right.

course/page/code.py
inducer: Thanks! That said, I think it's a risky assumption to say that this'll only be used for development. The simple fact is that this makes it possible to run Relate on Windows and OSX, and we should plan for folks using it that way--and thus make it bulletproof enough to be used in that setting. As such, it should aspire to using reasonable security practices, and `assert_hostname=False` doesn't strike me as one. It can be in there, commented out by default with an explanation, but it shouldn't be the default.

inducer:
There's actually already `RELATE_DOCKER_URL`.

dzhuang:
Sorry, I am buried with works. I will try to work on this in a few days.

inducer:
Closing in favor of https://gitlab.tiker.net/inducer/relate/merge_requests/39

PR #217, opened 2016-08-29
Require confirmation when submitting sessions with question(s) unanswered.

dzhuang:


course/templates/course/flow-confirm-completion.html
```diff
@@ -48,10 +60,46 @@
...
+        <button id="submit" type="submit" name="submit" class="btn btn-danger"
+                style="display:none"
+                data-content='<p>{% trans "Are you sure you will submit this session with questions unanswered?" %}</p>
+                        <button id="cancel-button" type="reset" name="cancel" class="btn btn-primary btn-sm" style="float: right;">
+                        {% trans "Cancel" %}</button>'
+                data-html="true"
+                data-container="body"
+                data-placement="right"
+          >
+            {# Translators: student should confirm to start another session. #}
```
inducer: This comment seems out of place.

inducer:
One wrinkle: Will the button still work even if someone has Javascript disabled?

I like this. Good to go once the two comments are addressed.

dzhuang:
I've removed the comment copied from the template ``flow-start`` :-). 

For your comment on Javascript: no, it won't work, just as the case when user need to start another flow session which might cause credit loss. Can you suggest any work around?

inducer:
Aww, whatever. Everyone has Javascript. :)

Thanks for the contribution!

PR #221, opened 2016-09-02
Continue: Remove use of RequestContext for Django 1.10 

### LINKED ISSUE #220: csrf_token missing for grading form
The rendered form does not have csrf_token.
```
<form  class="form-horizontal relate-grading-form" method="post" > <div id="div_id_grade_percent" class="form-group"> <label for="id_grade_percent" class="control-label col-lg-2">
                Grade percent ...
</form>
```

The server starts with warning:

```
C:\Python27\lib\site-packages\django\template\defaulttags.py:65: UserWarning: A
{% csrf_token %} was used in a template, but the context did not provide the val
ue.  This is usually caused by not using RequestContext.
  "A {% csrf_token %} was used in a template, but the context "
```

and as a result, the post of the grading form result in failure.

```
Reason given for failure:
    CSRF token missing or incorrect.
```

I'm using Django 1.10 and the repo is up-to-date, and I just can not figure out what is the cause.

dzhuang:
Continued for fd04b8fba32cc, should fix #220 .

### DIFF
--- a/course/page/base.py
+++ b/course/page/base.py
@@ -829,11 +829,9 @@ class PageBaseWithHumanTextFeedback(PageBase):
                 "rubric": markup_to_html(page_context, self.page_desc.rubric)
                 }
 
-        from django.template import RequestContext
         from django.template.loader import render_to_string
         return render_to_string(
-                "course/human-feedback-form.html",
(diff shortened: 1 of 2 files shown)


inducer:
Thanks!

PR #230, opened 2016-09-19
add param page_context to page.make_page_data

### LINKED ISSUE #213: Extending APIs for pages
1. Sometimes it is needed to include ``page_context`` into ``page.make_page_data`` method. For example, I want to load a repo data file with a number of different questions, and ask student to do one of them randomly. In that case, I want to store the question he/she was provided in page_data. However, to load the repo file, I need to use the ``page_context`` data.

1. Include ``request`` parameter in ``page.correct_answer`` method, so that I can provide an abstract answer for students while provide a detailed answer for TAs.

Are those ideas plausible? (I know how to do it.)

dzhuang:
Fix #213

### DIFF
--- a/course/flow.py
+++ b/course/flow.py
@@ -169,7 +169,7 @@ def _adjust_flow_session_page_data_inner(repo, flow_session,
         def create_fpd(new_page_desc):
             page = instantiate_page(new_page_desc)
 
-            data = page.make_page_data()
+            data = page.initialize_page_data(pctx)
             return FlowPageData(
                     flow_session=flow_session,
                     ordinal=None,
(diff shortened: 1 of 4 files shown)


inducer:
We can't really change the page API like that (since folks may have `Page` classes in their course content that we can't fix). Instead, you should add a `make_page_context_v2` entrypoint and have the default implementation spew a deprecation warning and fall back to the original `make_page_context` entrypoint.

dzhuang:
What name would you suggest for that new method? ``make_contextified_page_data`` or ``make_page_data_with_context`` or ``make_page_context_data``?

dzhuang:
Thanks for the precious suggestions!

inducer:
Looks good. Could you also change the built-in page types to use the new interface?

dzhuang:
I've change the implementation of builtin Choice question to that method. Do you mean I need to explicit call that method in all base page class?

inducer:
Oh, nvm. Choice is the only page type that actually makes use of this. Nvm, all good. Thanks for the patch!

PR #231, opened 2016-09-19
Fix #199 (continued).

### LINKED ISSUE #199: FlowPageVisit bug?

dzhuang:
Previous PR #202 for #199 only fixed in terms of GET request.

### DIFF
--- a/course/flow.py
+++ b/course/flow.py
@@ -1612,9 +1612,7 @@ def add_buttons_to_form(form, fpctx, flow_session, permissions):
     return form
 
 
-@retry_transaction_decorator()
-def create_flow_page_visit(request, flow_session, page_data,
-        answer_data, is_submitted_answer):
+def create_flow_page_visit(request, flow_session, page_data):
     # type: (http.HttpRequest, FlowSession, FlowPageData) -> None
 
     if request.user.is_authenticated:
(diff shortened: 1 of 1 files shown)


course/flow.py
```diff
@@ -1612,7 +1612,9 @@ def add_buttons_to_form(form, fpctx, flow_session, permissions):
     return form
 
 
-def create_flow_page_visit(request, flow_session, page_data):
+@retry_transaction_decorator()
+def create_flow_page_visit(request, flow_session, page_data,
+        answer_data, is_submitted_answer):
```
inducer: I'm not in principle against centralizing `FlowPageVisit` creation, but the way it's written forces even the logging of (fairly expendable) `GET` visits into a (hugely expensive, retry-hardened) `SERIALIZABLE` database transaction. The `GET` path should stay non-transactional. Maybe move split the function out into `create_flow_page_visit_transactional` and `create_flow_page_visit_base`?

dzhuang: Thanks for the informative suggestions. The simplest method is to revert it back and add 2 lines,  what about my latest push.

inducer: Works for me.

Issue #237, opened 2016-09-23
More fields on course model on email settings


dzhuang:
My idea is to include setup for the following functionalities after course creation:
   - allow email cc setting
   - allow reply-to in email header
for receive or not, receiver include TAs or not, for each kind of email, including registration, participation approval, students' question on human graded question after grades release  and TAs' question about grading of specific session. The last two need to be done by extending page api.

inducer:
Reply-to sounds good. The other things could also be permissions in the new permissions system. "Receives course email" etc.

Issue #195, opened 2016-04-02
Extending email setting


dzhuang:
I think there might be necessity to extend the settings of email configuration both at project level and course level. My initial idea is:

 1. at project level, add the following in local_settings.py:
      1. whether the email service used allow sending emails with different domain name, as discussed in #41 
      1. optionally a no-reply email addr for register/reset passwd. If none, use robot email.

 1. at course level
      1. whether TAs will receive email, and which kinds of email they will receive.
      1. reply-to settings of different events, who will receive response from participants regarding enroll decision/grading feedback/etc.

inducer:
Sure, that sounds good. I'd be happy to see an implementation of that.

dzhuang:
I think I need some advice.

1. May I parse the ``from_email`` and ``notify_email`` from course.yml rather than by saving a ``course.models.course`` instance from ``course.versioning.CourseCreationForm``? So that I don't need to manually input those values and I can select from a queryset which include the emails of tutor and TA in admin. 

1. I think at flow/page level, email might also need some configurations. 
    1. Sometimes the tutor is responsible for reviewing/grading/responding to some difficult flows (with human graded questions), TAs are responsible for other simpler flows. To achieve that, I think of adding a ``allowed_attribute`` named ``flow_reviewer``, which is default to ``course_reviewer`` if not defined, where course_reviewer can be set to TA or tutor.  
    1. For a single flow, there can also be difficult questions which require the review of tutor, and simple ones which require TAs to do the review. To achieve that, we can add a ``allowed_attribute`` named ``page_reviewer``, which is default to ``flow_reviewer`` if not defined.
    1. If the above settings are ok, response of grading feedback from participants will be sent to those reviews, and we can also extend some functionality like informing TA's to review flows and pages by automatic emails.

Can you provide some more advices, or pros and cons of my idea?

inducer:
1) I moved those from .yml content to the database on purpose, so I'd rather not see them go back, since they're more "daily operations" than "course content", and so I feel like they belong where they are. We could have them autocomplete from the user list though--as long as we allow custom entries, too.

2.i) I'm thinking of redoing the course permission/role system at some point, where the set "instructor"/"ta"/... becomes data rather than hard-coded concept, with a number of permissions attached. We could then add a "grading" section to the rules. This would provide an answer for #170 along with solving your use case, I thnk.

2.ii) There's already a notion of per-page rules--this would just need to be extended to the grading rules.

[...]

Issue #41, opened 2015-07-27
Issue when a tutor is sending emails


dzhuang:
If the tutor's email address (in the course configuration page) is different with ``SERVER_EMAIL`` in ``settings.py``,  sending an email by the tutor (e.g., sending enrollment decision) will give an error ``mail from address must be same as authorization user``. 

I have no idea whether this is the problem with my email service, or should there be some settings that the tutor should configure his email service.

dzhuang:
See http://stackoverflow.com/questions/20293986/how-do-i-send-email-from-a-users-email-address-with-django

It seems to be technically difficult to enable multiple tutors with different email address as sender on the website.

Maybe we can set a default email address to send notifications to users, and for some specific emails, e.g., enrollment decision, or feedback of exercise, add the tutors' email as the ``reply-to`` header so that students can communicate with tutor directly.

inducer:
For my courses, I made a policy decision that the students should not be aware who graded their work. But it would certainly be possible to include a 'cc me' box (or a corresponding course-wide setting). I'd be happy to take a patch.

dzhuang:
So do you mean you agree that ``course.from_email`` can not be set as a sender if it is not the same as the ``server_email``?

inducer:
No, I don't think I'm saying that--it should be possible to set it to anything you like.

dzhuang:
When I use different email addresses (``course.from_email="foo@domain1.com"``, settings.SERVER_EMAIL="bar@domain2.com``), I got an internal server error ``mail from address must be same as authorization user`` when I send an enrollment decision email to students. What the problem?

dzhuang:
Maybe it's the problem of my mail service provider.

inducer:
I'd imagine so.

dzhuang:
Thanks.

dzhuang:
The issue was solved when I use gmail as the site email. Now I have to use email other than gmail, as gmail is 404 in some country where I set up my server. I realize that most email service here do not allow sender to have a different email than the site email. There should be an option in the settings so that . I'll try to make a patch.

inducer:
Sounds good, will be happy to take it.

PR #238, opened 2016-09-24
Extending email settings

dzhuang:
1. Allow using email services which don't allow non-authorized sender, fixing #41 and #195
1. Allow multiple smtp, inspired by https://gist.github.com/niran/840999
1. Preparing for #237.

### DIFF
--- a/course/auth.py
+++ b/course/auth.py
@@ -435,13 +435,21 @@ def sign_up(request):
                         reverse("relate-home"))
                     })
 
-                from django.core.mail import send_mail
-                send_mail(
+                from django.core.mail import EmailMessage
+                msg = EmailMessage(
                         string_concat("[", _("RELATE"), "] ",
                                      _("Verify your email")),
                         message,
-                        settings.ROBOT_EMAIL_FROM,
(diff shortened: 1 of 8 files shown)


inducer:
Looks good. Thanks for the patch.

inducer:
FWIW, to avoid mistakes, Relate now gets tested with [mypy](http://mypy-lang.org) on every commit, and it's expected to pass. See `run-mypy.sh`. This commit broke that--see f53336ec for a fix. I've also renamed `get_connection` to `get_outbound_mail_connection`.

inducer:
This also broke the syntax in `local_settings.py.example`:
```
  File "local_settings.py", line 143
    "student_interact": {
                     ^
SyntaxError: invalid syntax
```

Please be more careful.

dzhuang:
Thank's for the information, I'll try to do the test before submitting a PR in the future, and sorry for the errors.

inducer:
Np. Thanks for submitting patches!

inducer:
d8644cb has more fixes for this that broke for me in production.

PR #240, opened 2016-09-25
Redirecting to sign out confirmation page for authenticated user to visit auth pages.

dzhuang:
Users who already logged in should not be able to visit sign in/sign up/reset password pages.


inducer:

PR #242, opened 2016-09-28
add warning if value is zero and no 'atol' for FloatMatcher.

dzhuang:


inducer:

PR #245, opened 2016-09-28
for py3.x, urljoin is in urllib.parse

dzhuang:


inducer:
Can you use `six` for that?

http://pythonhosted.org/six/#module-six.moves.urllib.parse

dzhuang:
Thanks for the information!

dzhuang:
Updated.

inducer:
Just import `six` directly, not via Django. (Sorry to be going on about this minor point.)

dzhuang:
Nvm. I've originally import it directly from six, and the switched to Django after I noticed some part of your code is importing from django :-). Modified to ``six`` now.

inducer:
Thx!

PR #246, opened 2016-10-03
Adding a new flow_permission: send_email_about_flow_page

dzhuang:
Preparing for #237

### DIFF
--- a/course/constants.py
+++ b/course/constants.py
@@ -356,6 +356,11 @@ class flow_permission:  # noqa
 
         (Optional) Once any page of the flow has been viewed, access to all content
         except for this session on this RELATE instance will be denied.
+
+    .. attribute:: send_email_about_flow_page
+
+        (Optional) If present, the participant can send interaction emails to
+        course staffs for questions for each page with that permission.
     """
     view = "view"
     end_session = "end_session"
@@ -368,6 +373,7 @@ class flow_permission:  # noqa
     set_roll_over_expiration_mode = "set_roll_over_expiration_mode"
     see_session_time = "see_session_time"
     lock_down_as_exam_session = "lock_down_as_exam_session"
+    send_email_about_flow_page = "send_email_about_flow_page"
 
 FLOW_PERMISSION_CHOICES = (
         (flow_permission.view,
@@ -397,6 +403,9 @@ FLOW_PERMISSION_CHOICES = (
             pgettext_lazy("Flow permission", "See session time")),
         (flow_permission.lock_down_as_exam_session,
             pgettext_lazy("Flow permission", "Lock down as exam session")),
+        (flow_permission.send_email_about_flow_page,
+         pgettext_lazy("Flow permission",
+                       "Send emails about the flow page to course staff")),
         )
 
 # }}}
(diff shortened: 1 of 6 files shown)


relate/urls.py
```diff
@@ -391,6 +391,15 @@
         "/$",
         course.flow.finish_flow_session_view,
         name="relate-finish_flow_session_view"),
+    url(r"^course"
+        "/" + COURSE_ID_REGEX +
+        "/flow-session"
+        "/(?P<flow_session_id>[0-9]+)"
+        "/(?P<ordinal>[0-9]+)"
+        "/page_interaction_email"
```
inducer: Can you make that `page-interaction-email`? Just because all the other URL bits use dashes.

course/constants.py
```diff
@@ -368,6 +368,8 @@ class flow_permission:  # noqa
     set_roll_over_expiration_mode = "set_roll_over_expiration_mode"
     see_session_time = "see_session_time"
     lock_down_as_exam_session = "lock_down_as_exam_session"
+    send_interact_email_to_course_staff =\
+        "send_interact_email_to_course_staff"
```
inducer: Not sure I get why this needs to be per-page, rather than one of the (new) general `participation_permission`s?

dzhuang: There should also be ones in the participation_permission(s), I will do that latter. The permission here is to restrict the usage of pages, I don't want to receive emails from students for all questions for all the time.

relate/urls.py
dzhuang: ok

course/constants.py
inducer: Please also add that to the docs for `flow_permission`.

dzhuang: Is that ok? I wish you can polish the naming of the code in order not to cause misunderstanding. Sorry for my bad English.

inducer: How about `send_email_about_flow_page`?

dzhuang: Thanks, updated.

course/templates/course/flow-page.html
```diff
@@ -284,6 +284,55 @@
...
+          >
+              {% trans "Grant exception" %}
+          </a>
+
+      <a class="btn btn-primary btn-xs relate-btn-xs-vert-spaced"
+         href="{% url 'relate-grade_flow_page' course.identifier flow_session.id ordinal %}"
+         role="button"
+         target="_blank"
+      >
+        {% trans "View in grading page" %}
```
inducer: Sorry that these comments come out so piece-by-piece, but I only just discovered (ahem) this bit of the PR. :)

The permissions for this need to be a bit finer-grained.

* Reopen is `pperm.reopen_flow_session`.
* Grant exception is `pperm.grant_exception`.
* Seeing the grading page is `pperm.view_gradebook`.

You actually don't need to pass a flag from Python to the template. Just

```
{% if pperm.XYZ %}
{% endif %}
```
should do the trick.

dzhuang: Never say sorry about that. I really like the opportunity to improve the code. Thanks for the information.   Updated.

inducer:
Looks good. Thanks!

PR #248, opened 2016-10-11
Default RELATE_EMAIL_SMTP_ALLOW_NONAUTHORIZED_SENDER to True 

dzhuang:
Default RELATE_EMAIL_SMTP_ALLOW_NONAUTHORIZED_SENDER to True if not configured in local_settings.py, so as to avoid inconsistency.

### DIFF
--- a/relate/settings.py
+++ b/relate/settings.py
@@ -18,6 +18,8 @@ import os
 from os.path import join
 BASE_DIR = os.path.dirname(os.path.dirname(__file__))
 
+RELATE_EMAIL_SMTP_ALLOW_NONAUTHORIZED_SENDER = True
+
 _local_settings_file = join(BASE_DIR, "local_settings.py")
 local_settings = {
         "__file__": _local_settings_file,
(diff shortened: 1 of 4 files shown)


relate/settings.py
```diff
@@ -287,4 +287,10 @@
     assert isinstance(
         local_settings["RELATE_EMAIL_APPELATION_PRIORITY_LIST"], list)
 
+# This defaults RELATE_EMAIL_SMTP_ALLOW_NONAUTHORIZED_SENDER to True if
+# not configured in local_settings.py
+RELATE_EMAIL_SMTP_ALLOW_NONAUTHORIZED_SENDER = getattr(
+        local_settings, "RELATE_EMAIL_SMTP_ALLOW_NONAUTHORIZED_SENDER",
+        True)
+
```
inducer: Why not just set that variable to `True` before the local_settings stuff gets set?

dzhuang: You're right :-), updated.

inducer:
Looks good. Thanks for taking care of this!

PR #252, opened 2016-10-14
Fine-tune flow finish button.

dzhuang:
Redefine ``noninteractive`` session kind, and then differentiate finish button icon/title between
- in-progress sessions and finished sessions.
- whether a session is expecting grade.

### DIFF
--- a/course/flow.py
+++ b/course/flow.py
@@ -594,9 +594,6 @@ def get_interaction_kind(
         ):
     # type: (...) -> Text
 
-    if not flow_session.in_progress:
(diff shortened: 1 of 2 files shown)


inducer:
This is fantastic. Thanks for putting this together.

Minor wrinkle (by which I pretend to understand something about accessibility on the web, which I don't): All the FontAwesome bits should probably have  `aria-hidden="true"` bits to it, too.

dzhuang:
Thanks for the info. Updated.

inducer:
Thx!

PR #253, opened 2016-10-17
update zh_CN translations.

dzhuang:
And let ``zh_HANs`` fall back to ``zh_CN``. Because from Django 1.9.6, ``zh_CN`` has been removed, while some dependencies (e.g., js packages) don't have a ``zh_HANs`` version.


inducer:
Thanks!

PR #254, opened 2016-10-17
More info for the form when send_email_about_flow_page.

dzhuang:
Warning only questions for that specific page is expected to be answered.


inducer:
Thx!

PR #255, opened 2016-10-17
Add page_data param in None API methods  in in InlineMultiQuestion

dzhuang:
I am wondering if these two methods (not api method) need to fall back to their original definition if I want to modify them? I need to have ``page_data`` for my customized pages.


inducer:
Nobody promised that `InlineMultiQuestion` would be safe to inherit from or API-stable, so just adding those is fine.

dzhuang:
thanks!

Issue #216, opened 2016-08-26
Cannot_see_flow_result permission not applying to gradebook


eecarrier:
The cannot_see_flow_result permission should make it so that the score doesn't appear in the student gradebook.  It'd be fine for the line to still appear in the student gradebook, just instead of score have the "no result shown".  This would eliminate the need to remember to hide/unhide the whole entry in the student gradebook and seems like it would be less confusing to the students.

inducer:
That's actually tricky/expensive to do, since Relate would have to evaluate flow permissions for each grade it shows. It's easy to add a "show/hide  results" knob in addition to the "show/hide row". It's also easy to make  the "edit grading opportunity" form more easily accessible from the gradebook. Would that help?

eecarrier:
Both of those would help, but it's not that big of a deal/easiest enough to manage if you don't have the time.  Just thought it was worth mentioning in case it was easy.

inducer:
The code to do this is now there. It'll go live at UIUC with a bigger batch of changes sometime soon.

inducer:
Now live.

PR #256, opened 2016-10-19
Use flow_desc to control display/hide of a gopp for participants

dzhuang:
Under current mechanism, a gopp is created if I test the flow in my account. That will let my student see it in their grade book. Sometimes I don't want student to see some gopp that are supposed to start later (like a month later) and kept answering their question "when should I submit that assignment" or "when can I start that session". 

So in this patch, I am using flow_desc to control whether a gopp should be displayed in participant's grade book. I know this patch is somewhat dirty, because the permission added is not for a session, but a gopp.  However, the benefit is that I don't need to manually set whether those flows should be display in participants' gradebook. Or, is there a better way to do that?

### DIFF
--- a/course/constants.py
+++ b/course/constants.py
@@ -361,6 +361,16 @@ class flow_permission:  # noqa
 
         (Optional) If present, the participant can send interaction emails to
         course staffs for questions for each page with that permission.
+
+    .. attribute:: cannot_see_in_participant_grade_book
+
+        (Optional) If present, participants won't see this flow in their
+        grade book. It can be used to hide grading opportunities.
+
+    .. attribute:: cannot_see_result_in_participant_grade_book
+
+        (Optional) If present, participants won't see the result of this 
+        flow in their grade book and single grade of this flow.
     """
     view = "view"
     end_session = "end_session"
@@ -374,6 +384,10 @@ class flow_permission:  # noqa
     see_session_time = "see_session_time"
     lock_down_as_exam_session = "lock_down_as_exam_session"
     send_email_about_flow_page = "send_email_about_flow_page"
+    cannot_see_in_participant_grade_book =\
+            "cannot_see_in_participant_grade_book"
+    cannot_see_result_in_participant_grade_book =\
+            "cannot_see_result_in_participant_grade_book"
 
 FLOW_PERMISSION_CHOICES = (
         (flow_permission.view,
@@ -406,6 +420,12 @@ FLOW_PERMISSION_CHOICES = (
         (flow_permission.send_email_about_flow_page,
          pgettext_lazy("Flow permission",
                        "Send emails about the flow page to course staff")),
+        (flow_permission.cannot_see_in_participant_grade_book,
+         pgettext_lazy("Flow permission",
+                       "Cannot see in participant grade book")),
+        (flow_permission.cannot_see_result_in_participant_grade_book,
+         pgettext_lazy("Flow permission",
+                       "Cannot see result in participant grade book")),
         )
 
 # }}}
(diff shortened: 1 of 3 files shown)


inducer:
This is a good idea. One thing I would like is for this to also be able to hide the results. (In the grade book, there are three states currently--show, hide results, and hide outright.) "Hide results" got added because students were confused about grades that had completely disappeared.

dzhuang:
Updated.

inducer:
Hum, now that I think of it, when @eecarrier asked for the same thing in #216, I turned it down based on the performance hit from having to materialize the `flow_desc` for each grade, if there is one.

I realize the utility of being able to do this, and with two of you asking for it, it's clear there's a need. Let me give this some thought.

inducer:
To get some data, how slowly does the "My grades" page load with this patch applied, for a student with, say, 30 grades?

dzhuang:
Will ``get_flow_desc_cached``, a cached version of ``get_flow_desc``,  help in increasing the performance?

dzhuang:
The difference between whether the patch is applied:

- before patched: 0.03 second
- after patched: 0.6 second

inducer:
0.2 seconds is borderline acceptable, 0.6 seconds is starting to feel kind of slow--especially since this is data that's (likely) from a largely idle system, so on a busy installation (~1000 active users), this has the potential to stink. Instantiating a new session is not really correct/consistent, so I'm not too happy with that as a solution.

This is already using `get_flow_desc_cached`, right? Can you try grabbing the access rules with exceptions turned off, just to see how much work goes into those queries?

inducer:
Forgot to say: Thanks for gathering that data!

dzhuang:
It's around 0.5-0.6 seconds if I grab the access rules with exceptions turned off. Here comes another problem, if a participant don't have a session for an opp, then how to determine whether the opp should be displayed in his/her gradebook? And a stupid question, why instantiate a session, which won't be saved, is not correct/consistent?

dzhuang:
In the past, we can use ``participation__role__iexact=pctx.participation.role`` to filter FlowSession objects, so we can filter the sessions that are not created by the user to grab the access_rules, if the user didn't have a session of that opp. But that can't be done since the role field is deprecated.

dzhuang:
Forgot to address ``get_flow_desc_cached``. I didn't create that method, as ``get_flow_desc`` returns a ``FlowDesc`` object, which is not cacheable directly. However, ``get_flow_desc`` is actually using cache by ``get_yaml_from_repo`` and ``markup_to_html``, with ``normalize_flow_desc`` not using cache. Maybe we can manage to optimize it. Is that necessary?

dzhuang:
I think I now understand why you say instantiate a session will cause incorrectness/inconsistency. Because during the period when that instance is used, the ``flow_desc`` of that opp might have been changed. Is that correct?

inducer:
> I think I now might partly understand why you say instantiate a session will cause incorrectness/inconsistency. Because during the period when that instance is used, the flow_desc of that opp might have been changed. Is that correct?

No, not really. it's more that a different set of grading rules may apply to the new instance than the actually referenced one. Worse, there's a potential contradiction here: What would happen if (say) multiple flow sessions contribute to a grade and they contradict each other on whether that gopp should be visible? Sure, one could err on the conservative side and only show the grade if everybody agrees it should be shown. But then hunting down the reason for why a grade isn't shown could turn into a wild goose chase.

So, between the performance and the design issues above--I'm just kind of not a fan of this.

dzhuang:
Thanks for the explanation, perhaps we need to find another workaround for this.

PR #258, opened 2016-10-25
Fix relative-error formula in feedback

bowmanat:


inducer:
I need to rebuild the container images in order to put this live. Will let you know once that's done.

inducer:
Done.

Issue #49, opened 2015-08-18
Inline questions do not render properly on mobile devices


inducer:
Leftover from #48

dzhuang:
Not a bug. This is the result with bootstrap inline forms, that a punctuation (comma, period or question mark) following a form field will look like leading a new paragraph, on mobile devices with screen width no more than 767px. 

To remove the unwanted punctuation from the form, we can wrap those punctuations with some element tag which won't be shown for small devices. I'll add docs to ``InlineMultiQuestion`` latter.

PR #262, opened 2016-11-07
Update inline functionality and fix errors.

dzhuang:
- Fix error when adding new embedded question to flow pages when there exist submitted answers for the page.
- Allow Text input area with prepended/appended text
- Remove feedback style when answer is changed for pages that allow user to change answer after see correctness
- Responsive for mobile view

### DIFF
--- a/course/page/inline.py
+++ b/course/page/inline.py
@@ -44,6 +44,7 @@ import re
 # {{{ multiple text question
 
 from crispy_forms.layout import Layout, Field, HTML
+from crispy_forms.bootstrap import PrependedAppendedText
 
 
 class InlineMultiQuestionForm(StyledInlineForm):
@@ -174,23 +175,19 @@ class AnswerBase(object):
             return 0
 
     def get_field_layout(self, correctness=None):
+        kwargs = {}
+        kwargs["prepended_text"] = getattr(self.answers_desc, "prepended_text", "")
+        kwargs["appended_text"] = getattr(self.answers_desc, "appended_text", "")
+        kwargs["use_popover"] = "true"
+        kwargs["popover_title"] = getattr(self.answers_desc, "hint_title", "")
+        kwargs["popover_content"] = getattr(self.answers_desc, "hint", "")
         if correctness is None:
-            return Field(
-                    self.name,
-                    use_popover="true",
-                    popover_title=getattr(self.answers_desc, "hint_title", ""),
-                    popover_content=getattr(self.answers_desc, "hint", ""),
-                    style=self.get_width_str()
-                    )
+            kwargs["style"] = self.get_width_str()
         else:
-            return Field(
-                    self.name,
-                    use_popover="true",
-                    popover_title=getattr(self.answers_desc, "hint_title", ""),
-                    popover_content=getattr(self.answers_desc, "hint", ""),
-                    style=self.get_width_str(self.width + 2),
-                    correctness=correctness
-                    )
+            kwargs["style"] = self.get_width_str(self.width + 2)
+            kwargs["correctness"] = correctness
+
+        return PrependedAppendedText(self.name, **kwargs)
 
     def get_form_field(self, page_context):
         raise NotImplementedError()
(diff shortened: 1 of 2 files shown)


inducer:
What's the use case for prepended/appended text?

course/page/inline.py
```diff
@@ -543,17 +550,20 @@ class InlineMultiQuestion(TextQuestionBase, PageBaseWithValue):
...
-            # An example
+            # An InlineMultiQuestion example
 
             Complete the following paragraph.
 
         question: |
 
             Foo and [[blank1]] are often used in code examples, or
-            tutorials. The float value of $\frac{1}{5}$ is [[blank_2]].
+            tutorials. $\frac{1}{5}$ is equivelant to [[blank_2]].
```
inducer: Typo.

course/page/inline.py
```diff
@@ -773,9 +816,15 @@ def make_form(self, page_context, page_data, answer_data, page_behavior):
...
-                    if answer[answer_instance.name] is not None:
-                        correctness_list.append(answer_instance.get_correctness(
-                                answer[answer_instance.name]))
+                    try:
+                        if answer[answer_instance.name] is not None:
+                            correctness_list.append(answer_instance.get_correctness(
+                                    answer[answer_instance.name]))
+
+                    # The answer doesn't exist for newly added question 
+                    # for pages which have been submit.
```
inducer: submitted.

course/page/inline.py
```diff
@@ -174,23 +175,19 @@ def get_weight(self, answer):
...
-                    use_popover="true",
-                    popover_title=getattr(self.answers_desc, "hint_title", ""),
-                    popover_content=getattr(self.answers_desc, "hint", ""),
-                    style=self.get_width_str(self.width + 2),
-                    correctness=correctness
-                    )
+            kwargs["style"] = self.get_width_str(self.width + 2)
+            kwargs["correctness"] = correctness
+
+        return PrependedAppendedText(self.name, **kwargs)
```
inducer: Does this work OK even if nothing is prepended/appended?

dzhuang: Yes, according to the definition of ``PrependedAppendedText``, it is a subclass of Field with ``prepended_text/appended_text=None`` as optional arguements.
```
class PrependedAppendedText(Field):
    template = "%s/layout/prepended_appended_text.html"

    def __init__(self, field, prepended_text=None, appended_text=None, *args, **kwargs):
        self.field = field
        self.appended_text = appended_text
        self.prepended_text = prepended_text
        if 'active' in kwargs:
            self.active = kwargs.pop('active')

        self.input_size = None
        css_class = kwargs.get('css_class', '')
        if 'input-lg' in css_class:
            self.input_size = 'input-lg'
        if 'input-sm' in css_class:
            self.input_size = 'input-sm'

        super(PrependedAppendedText, self).__init__(field, *args, **kwargs)
```

What's more, I implemented that in the ``AnswerBase``  as I had thought that it can also be used for ``ChoicesAnswer``, and I did try to add ``prepended_text`` and ``appended_text`` also to the ``allowed_attr`` for ``ChoicesAnswer`` question, but I then removed them as I found it has been suggested not doing so by Bootstrap docs[http://getbootstrap.com/components/#input-groups]:
>
## Textual `<input>`s only
Avoid using `<select>` elements here as they cannot be fully styled in WebKit browsers.
Avoid using `<textarea>` elements here as their rows attribute will not be respected in some cases.
>

Shall I revert the revision of  ``get_field_layout`` back in ``AnswerBase``  and make a implementation for ``ShortAnswer``? I felt it unnecessary according to the declaration of ``PrependedAppendedText``.

dzhuang:
As for 
> What's the use case for prepended/appended text?

It is primarily for the rendering problem of bootstrap inline-forms on small screens (max-width 768), which had also been addressed in #49 . The differences before and after the fix, on a mobile phone:

Before
before_implementation

After
implementation

As you can see, all the ``.``  and ``,`` after the inputs are put into a new line, that's tolerable for me, but for the newly added questions, that's too weird.

course/page/inline.py
inducer: No, I think that's fine.

inducer:

PR #264, opened 2016-11-14
Order GradeChange objects of a participation by flow_session.completion_time

### LINKED ISSUE #263: Bug: wrong overall grade when grading criterion changed between multiple submissions

- 10% at 9:00am
- 20% at 9:01am
- 30% at 9:02am

dzhuang:
Fix #263.


inducer:
Not every grade change has a flow session. (Think manually imported grades from outside Relate.) This would cause this code to creash.

dzhuang:
You are right, my fault.

PR #265, opened 2016-11-22
Add permission: cannot_see_answer_for_sessions_in_progress

dzhuang:
``see_answer_before_submission`` and ``see_answer_after_submission`` seems only control at the page level, not flow level. 

In some situations, I wish not to show correct answer to participants for in-progress sessions, with pages which allow them to see correctness after page submission, and allow them to change answer, so as to encourage them to figure out the correct answer themselves. This cannot be done by current permissions, with ``see_answer_after_submission``, they will be allowed to copy from the correct answer after page submission.

### DIFF
--- a/course/constants.py
+++ b/course/constants.py
@@ -361,6 +361,12 @@ class flow_permission:  # noqa
 
         (Optional) If present, the participant can send interaction emails to
         course staffs for questions for each page with that permission.
+
+    .. attribute:: cannot_see_answer_for_sessions_in_progress
+
+        (Optional) If present, shows the correct answer to the participant if only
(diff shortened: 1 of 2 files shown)


dzhuang:
UPDATE: The PR is not working, close now, and will reopen if fixed.

dzhuang:
Can you review it @inducer ? Thanks.

inducer:
I'm not liking that. You can just use `if_in_progress: True` and specify one set of rules and then use `if_in_progress: False` to specify another. So in that sense, this feature duplicates something that can already be done, which tends to be confusing.

dzhuang:
Opps, sorry for the foolish question, and thanks for the smart mechanism and your explanation.

PR #271, opened 2017-01-21
Use bleach to clean HTML in code feedback.

### LINKED ISSUE #267: More accurate categorization of error type in code question?
As the python code is lazily evaluated, sometimes a problematic code from user will pass ``exec``, and the error won't be raised until it is runned in ``test_code``. That bring me tons of email. I know we can define some try-exception to prevent that, and use some code to generate the error messages, but we need to put that code in every ``test_code`` which might encounter the problem. Is it possible to create some ``Exception`` types, when which is raised, the runpy will categorize the exception to ``user_error`` instead of ``test_error`` so that won't trigger the error email?

dzhuang:
Fix #267 .

### DIFF
--- a/course/page/code.py
+++ b/course/page/code.py
@@ -770,13 +770,21 @@ class PythonCodeQuestion(PageBaseWithTitle, PageBaseWithValue):
             raise RuntimeError("invalid runpy result: %s" % response.result)
 
         if hasattr(response, "feedback") and response.feedback:
+            def sanitize(s):
+                import bleach
+                return (
+                    bleach.clean(s, tags=["p", "pre"])
+                        # html5lib will turn <user code> to <user code="">
+                        # https://github.com/html5lib/html5lib-python/issues/194
+                        .replace("=\"\"&gt;", "&gt;")
+                )
             feedback_bits.append("".join([
                 "<p>",
                 _("Here is some feedback on your code"),
                 ":"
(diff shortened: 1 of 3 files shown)


course/page/code.py
```diff
@@ -770,13 +770,21 @@ def transfer_attr(name):
             raise RuntimeError("invalid runpy result: %s" % response.result)
 
         if hasattr(response, "feedback") and response.feedback:
+            def sanitize(s):
+                import bleach
+                return (
+                    bleach.clean(s, tags=["p", "pre"])
+                        # html5lib will turn <user code> to <user code="">
```
inducer: We could just switch to different delimiters there, e.g. `[user code]`.

course/page/code_feedback.py
```diff
@@ -166,7 +166,7 @@ def call_user(self, f, *args, **kwargs):
             from traceback import format_exc
             self.add_feedback(
                     "<p>"
-                    "The function '%s' supplied in your code failed with "
+                    "The '%s' supplied in your code failed with "
```
inducer: Maybe replace "function" with "callable"? I'm not sure no noun is better.

course/page/code.py
dzhuang: Thanks. Updated.

inducer:
LGTM. Thanks for the contribution. I'll do a container rebuild sometime soon.

PR #272, opened 2017-01-21
Fix display of form validation messages for inline questions.

### LINKED ISSUE #266: Inline questions failed to show form validation error messages.
The bug seems to be caused by [here](https://github.com/inducer/relate/blob/8d79aa1ca02beacaf4ec545ce6125efa5011e78a/course/templates/course/custom-crispy-inline-form.html#L48). I'll submit a PR to solve this.

dzhuang:
Fix #266 , since this patch has nothing to do with python, I submit it here rather than Gitlab.


inducer:
Thx!

PR #282, opened 2017-02-08
Optimize mobile view for grading interface.

dzhuang:


relate/static/css/style.css
```diff
@@ -202,26 +202,48 @@ div.progress:hover  span.stats-percentage{
...
-.grading-page-grade-entry {
-  position: absolute;
-  top: 60px;
-  right: 0;
-  width: 50%;
-  bottom: 0;
-  overflow-y: auto;
-  padding-left:50px;
-  padding-right:50px;
+@media (min-width: 767px) {
```
inducer: Where does the "767" come from? What about a viewport that's *exactly* 767px wide?

dzhuang: Sorry, that's my bad, modified to "768".

inducer:
Thanks for submitting this!

PR #285, opened 2017-02-10
Toleration for float point correctness

dzhuang:
When correctness value is result of float calculation, there should be toleration. For example, answer for a code question with correctness summed up may get 0.9999999998, which was expected to be 1, will result in "Your answer is mostly correct."


inducer:
Fair point. :)

PR #291, opened 2017-02-28
Add an ALWAYS_CORRECT tag for MultipleChoice questions

bowmanat:
I don't have a working RELATE instance to test this with (sorry), so please test it out.

I thought we might want some explanation near line 380 for how the new tag is different from DISREGARD and why it might be used, but I couldn't come up with something short and sweet and didn't want to clutter up the documentation.

I think the string_concat added at the end should do what we want, but I don't know if I wrote it the way you would typically use that function.  I just used the previous message as a model.

### DIFF
--- a/course/page/choice.py
+++ b/course/page/choice.py
@@ -70,6 +70,7 @@ def markup_to_html_plain(page_context, s):
 class ChoiceQuestionBase(PageBaseWithTitle, PageBaseWithValue):
     CORRECT_TAG = "~CORRECT~"
     DISREGARD_TAG = "~DISREGARD~"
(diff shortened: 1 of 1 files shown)


inducer:
Thanks! Looks good aside from the one minor fix I made. I'll update relate.cs tonight so that we can wrap up grading our exam.

> didn't want to clutter up the documentation.

Not really a huge concern, please feel free to PR something that illustrates why that might be needed.

inducer:
FWIW, next time please submit PRs at https://gitlab.tiker.net/inducer/relate--that way the tests/linters run automatically.

Issue #292, opened 2017-03-09
Shortest time flow can show up as out of bounds on time histogram due to rounding error


eecarrier:
With logarithmic histograms (like time histogram), start of 1st bin is computed as `exp(log(min_value))` which should theoretically evaluate to min_value.  But, in some cases due to rounding error it will evaluate to a number slightly larger than min_value (for example if min_value=10.0) which causes the shortest time flow to incorrectly show up as out of bounds.

inducer:
Thanks for the report! Could you submit a patch/pull request?

inducer:
Thx!

PR #293, opened 2017-03-10
Fix log histogram out of bounds bug (#292)

eecarrier:
Couldn't test as I don't have a working Relate installation.


inducer:
Low risk--all that'll get broken is analytics. :) Thanks for providing this fix.

Now live on relate.cs. Please try and report bakc.

eecarrier:
All looks good, this fixed the out of bounds bug I was seeing.

PR #296, opened 2017-03-13
Fix evil subclasses hack

### LINKED ISSUE #295: Insecure type checking in code_feedback
It is easy to trick the autograder into accepting an object that cheats the correctness checks. Here's two examples that bypass the [scalar value check](https://github.com/inducer/relate/blob/master/course/page/code_feedback.py#L135), tested on CS357 quiz 17:

```
# Type not being checked when is_number attribute is True
# https://github.com/inducer/relate/blob/master/course/page/code_feedback.py#L144
class Evil:
    def __init__(self):
        self.is_number = True
    def __abs__(self):
        return self
    def __rsub__(self, o):
        return self
    def __lt__(self, o):
        return True

answer = Evil()
```

```
# Same idea, since we subclassed one of the accepted types
# isinstance() will let us through. Might need to change the
# superclass based on what the accepted answer type is, since
# __rsub__ doesn't get called if they're the same type.
class Evil2(np.float64):
    def __abs__(self):
        return self
    def __rsub__(self, o):
        return self
    def __lt__(self, o):
        return True

answer = Evil2()
```

Changing `isinstance(x, (...))` to `type(x) in [...]` (and fixing the `is_number` check) would prevent this particular trick by disallowing subclasses, though I'm not sure if that would be secure either.

Sorry if you don't want this code floating around in public, let me know and I'll remove it ASAP.

fschr:
closes #295

### DIFF
--- a/course/page/code_feedback.py
+++ b/course/page/code_feedback.py
@@ -135,16 +135,13 @@ class Feedback:
     def check_scalar(self, name, ref, data, accuracy_critical=True,
             rtol=1e-5, atol=1e-8, report_success=True, report_failure=True):
         import numpy as np
+        import sympy as sp
 
-        if not isinstance(data, (complex, float, int, np.number)):
-            try:
-                # Check whether data is a sympy number because sympy
-                # numbers do not follow the typical interface
-                # See https://github.com/inducer/relate/pull/284
-                if not data.is_number:
-                    self.finish(0, "'%s' is not a number" % name)
-            except AttributeError:
-                self.finish(0, "'%s' is not a number" % name)
+        if not type(data) in (complex, float, int, np.number,
+                              sp.Number, sp.Float, sp.Integer,
+                              sp.Rational, sp.AlgebraicNumber):
+            # the built-in `type` function cannot be fooled
+            self.finish(0, "'%s' is not a number" % name)
 
         good = False
 


course/page/code_feedback.py
```diff
@@ -135,16 +135,13 @@ def check_list(self, name, ref, data, entry_type=None):
     def check_scalar(self, name, ref, data, accuracy_critical=True,
             rtol=1e-5, atol=1e-8, report_success=True, report_failure=True):
         import numpy as np
+        import sympy as sp
```
inducer: I'd rather not. `import sympy` looks innocuous, but it takes about a second--so we implicitly add that overhead to every autograder submission, whether it uses sympy or not.

course/page/code_feedback.py
```diff
@@ -135,16 +135,13 @@ def check_list(self, name, ref, data, entry_type=None):
...
-        if not isinstance(data, (complex, float, int, np.number)):
-            try:
-                # Check whether data is a sympy number because sympy
-                # numbers do not follow the typical interface
-                # See https://github.com/inducer/relate/pull/284
-                if not data.is_number:
-                    self.finish(0, "'%s' is not a number" % name)
-            except AttributeError:
-                self.finish(0, "'%s' is not a number" % name)
+        if not type(data) in (complex, float, int, np.number,
```
fschr: `np.number` is an abstract base class, so this is actually incorrect.

course/page/code_feedback.py
apsun: It should be possible to use `sys.modules` to check if `sympy` was imported, and perform additional checks based on that. Something like this:

```
def check_type(data):
    import numpy as np

    # Built-in and numpy types
    if type(data) in (complex, float, int, np.float32, np.float64, np.int32, np.int64):
        return True

    # If the user imported sympy, also allow sympy types
    # This reuses the module that's already been imported
    if 'sympy' in sys.modules:
        sp = sys.modules['sympy']
        if type(data) in (sp.Number, sp.Float, sp.Integer, sp.Rational, sp.AlgebraicNumber):
            return True

    return False
```

inducer: I suggested that, but @bowmanat decided not to go this route in #284.

PR #312, opened 2017-04-09
Use --no-location to generate i18n .po files. Update zh_CN translation.

dzhuang:
To reduce the meaningless position diff lines when positions of i18n literal changed.
e.g.

```
python manage.py makemessages -l de --no-location
```

### DIFF
--- a/locale/de/LC_MESSAGES/django.po
+++ b/locale/de/LC_MESSAGES/django.po
@@ -8,31 +8,27 @@ msgid ""
 msgstr ""
 "Project-Id-Version: PACKAGE VERSION\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2016-02-19 13:57-0600\n"
+"POT-Creation-Date: 2017-04-09 23:42+0800\n"
 "PO-Revision-Date: YEAR-MO-DA HO:MI+ZONE\n"
 "Last-Translator: FULL NAME <EMAIL@ADDRESS>\n"
 "Language-Team: LANGUAGE <LL@li.org>\n"
-"Language: \n"
 "MIME-Version: 1.0\n"
 "Content-Type: text/plain; charset=UTF-8\n"
 "Content-Transfer-Encoding: 8bit\n"
+"Language: \n"
 "Plural-Forms: nplurals=2; plural=(n != 1);\n"
 
-#: accounts/apps.py:8
 #, fuzzy
 #| msgid "Counts"
 msgid "Accounts"
 msgstr "Anzahlen"
 
-#: accounts/models.py:42
 msgid "username"
 msgstr "Benutzername"
 
-#: accounts/models.py:46
 msgid "Required. 30 characters or fewer. Letters, digits and @/./+/-/_ only."
 msgstr ""
 
-#: accounts/models.py:50
 msgid ""
 "Enter a valid username. This value may contain only letters, numbers and @/./"
 "+/-/_ characters."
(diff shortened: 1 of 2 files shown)


dzhuang:
I think the idea not good, because if we update all .po files, the date information in all *.po files will be updated, that will cause unmodified files to generate a diff. 

I did try to write a script for user to select which locale they want to update, but it need to use bash, not sh, because I find now way to create a locale array, for user to select from, using sh .

We can just modify to doc to tell contributor if they want to contribute to the translation, then they need to add that parameter when submit the po files.

inducer:
Hmm, the more I think about it, the less I really like it--I find the location info sometimes helpful for context.

dzhuang:
You are right on that.  

However, the reason I want to remove location is very simple: there are too many diff lines each time we submit an updated po files. Most of those diff lines, sometimes more than 95%, are about the position changes of i18n literals changed when the source code is changed. That make it hard to find out whether there are new literals, and if there are, where those literals are which need to be translated.

As for context, those marked context in the form of ``pgettext`` or `` Translators`` or ``context`` will be kept in po files though applying ``--no-location``. 

If we use that ``--no-location`` parameter, the workflow (for an update) I suggest is: 

1. Generate a po file without location
1. Findout whether there're new literals by diff. iff yes, goto the next step, else stop
1. During translation, if there're problems about context, regenrate that po file without location and do the translation.
1. If the developer want to contribute to the project, submit the po files without location.

For contributions of a new language, the workflow is:
1. Generate a po file with location
1. Do the translation
1. Submit the po files without location.

inducer:
> Most of those diff lines, sometimes more than 95%, are about the position changes of i18n literals changed when the source code is changed. That make it hard to find out whether there are new literals, and if there are, where those literals are which need to be translated.

You've *almost* convinced me. One final concern: Most editors (Vim/Emacs) provide some support for editing `.po` files that makes common tasks like finding untranslated strings easy:

* http://www.vim.org/scripts/script.php?script_id=695
* https://www.emacswiki.org/emacs/PoMode

Do you think this would be adequate to convince you to keep the location info? If not, I'm happy to merge this.

dzhuang:
I had to say, not keeping location info seems to be a convention for i18n for , if not all python project, as least Django. You can see this [example](https://github.com/django/django/blob/3c447b108ac70757001171f7a4791f493880bf5b/django/contrib/auth/locale/de/LC_MESSAGES/django.po). 

One pro I think you might convince you, as a PR reviewer, for not keeping location info in repo: imaging a contributor modified your [de po file](https://github.com/inducer/relate/blob/master/locale/de/LC_MESSAGES/django.po). He did the ``makemessage`` but what he modified is just a tiny modification, such as removing one redundant space in the translation. However, because the po file had not been updated since a year before, most of the location info of literals are changed. Then you will find youself in trouble in reviewing the changes, in finding where the actual modification are, because the modifications are buried in diff lines of location changes. What make it worse, github refuse to display diff info in webpages because there were too much modificiations. If not keep the location info, reviewing changes will be *at a glance*.

Another con for keeping the location info: I often shame myself, as a contributor, that most of my contribution lines to the project, are those location change lines, which count up to tens of thousand :-(.

inducer:
Alright, you got me convinced. :)

inducer:
Thanks for taking the time to talk this through... ;)

dzhuang:
You are welcome. In fact, I felt (which you might also felt) myself improved so much these years in coding.  The reason is having the chance to contribute to the project, and learn from your code. Thanks!

PR #314, opened 2017-04-17
Allow submit buttons at bottom in human grading forms.

dzhuang:
This is necessary because the human grading form is fairly long.


inducer:
Thx!

PR #332, opened 2017-05-19
Deep clone for submit_and_next buttons

### LINKED ISSUE #331: Submit+next buttons at top in grading interface broken
(and I just confirmed in different browsers that the submit/nextpage and submit next session buttons do not work)

cc @lukeolson

dzhuang:
Fixed #331 , according to http://stackoverflow.com/a/9549716/3437454


inducer:
Thanks!

PR #337, opened 2017-06-20
Update static paths to match bower_components

dfrdmn:
From a fresh install, the paths in the css/js in the templates (`{% static '/path/to/css/or/js' %}`) doesn't match the bower installation.


inducer:
Thanks!

PR #338, opened 2017-06-23
Email field max_length should match User model.

dfrdmn:
Instead of hard-coding a length in `SignInByEmailForm`, grab the email field length from the model to enforce in the form.

It's not quite necessary in `ResetPasswordFormByEmail`, but probably doesn't hurt.


inducer:
Thx!

PR #339, opened 2017-07-03
Don't show 'View grades' navbar menu for anonymous user.

dfrdmn:
Unless I'm mistaken, this will always lead to a 403 error, so may as well just not show it.


inducer:
Thanks!

PR #341, opened 2017-07-03
List first name first in user profile form

dfrdmn:
While first name first is not culturally universal, this order matches the model and the Meta fields order in the form.

### DIFF
--- a/course/auth.py
+++ b/course/auth.py
@@ -835,7 +835,7 @@ class UserForm(StyledModelForm):
         super(UserForm, self).__init__(*args, **kwargs)
 
         self.helper.layout = Layout(
-                Div("last_name", "first_name", css_class="well"),
+                Div("first_name", "last_name", css_class="well"),
                 Div("institutional_id", css_class="well"),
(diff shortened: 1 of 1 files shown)


inducer:
@dzhuang, comments?

inducer:
(I don't mind either way.)

dzhuang:
Hi @dfrdmn . I think what you proposed can done by adding a setting (with name, for example ``RELATE_USER_PROFILE_FIRST_NAME_FIRST`` with default ``True`` ) in ``local_settings.py``, to let the site owner to determine the order of first_name and last_name in the profile form.

inducer:
@dfrdmn Do you feel strongly enough about this to make this a setting? I'm a bit reluctant but could be convinced.

dfrdmn:
I don't feel super strongly about this, so I'd argue to close this request without the merge. If, in the future, people find it and add to it, then adding a setting probably makes sense.

inducer:
Sounds good. Thanks for being flexible. I'll close this for now.

PR #342, opened 2017-07-03
Settings option to hide inst. ID and editor forms in user profile

### LINKED ISSUE #340: Feature: Hide Institutional ID and/or Editor forms in profile
We are setting up RELATE (awesome btw!) and have no need for the institutional ID or editor parts of the profile. As a quick fix to reduce confusion with users, we are simply hiding the layouts by creating two settings:

    RELATE_SHOW_INST_ID_FORM = False
    RELATE_SHOW_EDITOR_FORM = False 

We appended the `adjust_layout` function by adding to display/hide based on these settings:

    if not settings.RELATE_SHOW_INST_ID_FORM:
        self.helper.layout[1].css_class = 'well hidden'
    if not settings.RELATE_SHOW_EDITOR_FORM:
        self.helper.layout[2].css_class = 'well hidden'

This seemed like the most straightforward approach to ensure that no form validation breaks along the way. If you think this is worthwhile, I can create a pull requestor modify it per your needs and then create the request.

dfrdmn:
Closes #340

### DIFF
--- a/course/auth.py
+++ b/course/auth.py
@@ -859,6 +859,10 @@ class UserForm(StyledModelForm):
             else:
                 self.fields["institutional_id"].widget.\
                         attrs['disabled'] = True
+            if not settings.RELATE_SHOW_INST_ID_FORM:
+                self.helper.layout[1].css_class = 'well hidden'
+            if not settings.RELATE_SHOW_EDITOR_FORM:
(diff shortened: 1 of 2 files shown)


inducer:
Thanks!

inducer:
1fb12c6 is another defaulting fix, similar to the one for username login.

PR #345, opened 2017-07-12
Configuration for email-only signin.

dfrdmn:
We want a setup where users can _only_ use the email sign-in. This pull request mimics the other logic sign-in logic:

1. Add a `RELATE_SIGN_IN_BY_USERNAME_ENABLED` setting (default: True) which can be used to disable username sign-in in the app (but not /admin/).
2. Since there is not a guaranteed second sign-in method, check before showing the alert about alternate methods.

There may be some other logic that I'm missing (or a good reason to not disable the username sign-in). Happy to update and/or scrap depending on your thoughts.

### DIFF
--- a/course/auth.py
+++ b/course/auth.py
@@ -352,6 +352,11 @@ def sign_in_by_user_pw(request, redirect_field_name=REDIRECT_FIELD_NAME):
     """
     Displays the login form and handles the login action.
     """
+    if not settings.RELATE_SIGN_IN_BY_USERNAME_ENABLED:
+        messages.add_message(request, messages.ERROR,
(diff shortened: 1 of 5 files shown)


inducer:
Looks good. No objections that I can think of. Thanks!

inducer:
Hi @dfrdmn, I didn't catch this during review, but just FYI, I'm defaulting the setting to True in `relate/settings.py`. This avoids broken sites (like *cough* the one I run *cough*) on upgrade.

dfrdmn:
Haha, yes, if there is one deployment we probably shouldn't break it's yours!

PR #364, opened 2017-09-24
Fixed #363

### LINKED ISSUE #363: DEFAULT_FROM_EMAIL is undocumented
@dzhuang It looks like `Course.get_from_email` uses `settings.DEFAULT_FROM_EMAIL`, but that setting doesn't occur in either `relate/settings.py` or `local_settings.example.py`, so it's not very discoverable and likely to lead to errors when users need to set up From-restricted email.

Could you take a look?

dzhuang:
I think that will fix #363. I'll add unittests later.


inducer:
LGTM, thx!

PR #365, opened 2017-09-24
Optimize display of InlineMulti question.

dzhuang:
If there were literals before the first blank, those literal will be wrapped by `<p>`, displaying an indepedent paragraph, like this:

image

With this patch, it will check whether the first element is wrapped by `<p>`, if yes, modify the element to be wrapped by `<span>`, then :
image

This patch is kind of dirty, because I think it's hard to do by hacking the form template, and actually I didn't find where to hack that.


dzhuang:
Can not reproduce again, closed.

PR #366, opened 2017-09-26
Allow set mail (connection) for enrollment (request and decision).

dzhuang:
Sometimes we need to make those emails indepedent.


inducer:
What will this do if someone upgrades and doesn't have the `enroll` email setting?

dzhuang:
Then it will just work as before.

dzhuang:
Sorry I should have addd some tests.

inducer:
Looks good--let me know once you've got the tests in.

dzhuang:
Overrided by #371

PR #367, opened 2017-09-28
Facilitate local testing.

dzhuang:
This make it easier to do test locally, without overwrite `local_settings.py` (for production) each time we are doing tests. The default local_settings file will be `local_settings.example.py`. To override that default value, use `--local_test_settings` param.

The patch double secured local_settings.py won't be used for unittests. 

See `python manage.py test -h` for help.


inducer:
LGTM, thanks!

PR #368, opened 2017-09-29
Facilitate tests by adding __init__.py and exclude it from setup

dzhuang:
I've moved `check.py` to relate directory, because those checks are at the project level, not belong to course module, though it is called when Course.app is ready.


dzhuang:
I've move `check` to relate directory. It should be a project level module, not belong to course module, though it is called when Course.app is ready.

dzhuang:
The change of structure:
image

dzhuang:
This make it easier for write tests, e.g., make import, code auto-completion, and refactor more convinient.

inducer:
Any reason for the `test` -> `tests` rename?

inducer:
Looks fine otherwise.

dzhuang:
Sorry, I felt it a tradition to add "s" for test folder? Because I see many projects (including [Django repo](https://github.com/django/django))  were using tests for unit tests. If you mind it, I'll revert that.

inducer:
All my other projects use `test/`, but I'm flexible. The Django convention seems fine.

dzhuang:
> All my other projects use test/

Sorry, I didn't know that.

Issue #370, opened 2017-09-30
Is that a bug: user can send participation request (email) even when he had a pending request?


dzhuang:
Though there's no short cut in the page, he can post a request to do that.

inducer:
Yes, I'd consider that a bug.

dzhuang:
Closing, as this was fixed in 93b30d3.

PR #371, opened 2017-09-30
Fixed #369, #370. Overrided #366

### LINKED ISSUE #369: Is that a bug: unconfirmed users can request enrollment for courses in some cases
It happens when a course does not require email suffix, I found this when writing tests.

dzhuang:


course/enrollment.py
```diff
@@ -213,11 +270,12 @@ def enroll_view(request, course_identifier):
     if (
             preapproval is None
             and course.enrollment_required_email_suffix
-            and not user.email.endswith(course.enrollment_required_email_suffix)):
+            and not user.email.endswith(
+                "@%s" % course.enrollment_required_email_suffix.lstrip("@"))):
```
dzhuang: This fix both `@abcd.com` `@bcd.com`  match `bcd.com` as suffix.

course/enrollment.py
```diff
@@ -902,6 +972,13 @@ def __init__(self, add_new, pctx, *args, **kwargs):
 
         if not add_new:
             self.fields["user"].disabled = True
+        else:
+            participation_users = Participation.objects.filter(
+                course=participation.course).values_list("user__pk", flat=True)
+            self.fields["user"].queryset = (
+                get_user_model().objects.exclude(pk__in=participation_users)
+            )
```
dzhuang: With this, less chance to raise `IntegrityError`.

course/enrollment.py
```diff
@@ -937,6 +1014,16 @@ def __init__(self, add_new, pctx, *args, **kwargs):
...
 
+    def clean_user(self):
+        user = self.cleaned_data["user"]
+        if not self.add_new:
+            return user
+        if user.status == user_status.active:
+            return user
+
+        from django.forms import ValidationError as FormValidationError
+        raise FormValidationError(VALIDATION_ERROR_USER_NOT_CONFIRMED)
```
dzhuang: Avoiding adding users who are not `active`.

dzhuang:
I think the PR is ready for reviewed now.

course/enrollment.py
```diff
@@ -76,6 +76,49 @@
...
+    "Confirm your email to continue.")
+MESSAGE_PARTICIPATION_CHANGE_SAVED_TEXT = _("Changes saved.")
+
+EMAIL_NEW_ENROLLMENT_REQUEST_TITLE_PATTERN = (
+    string_concat("[%s] ", _("New enrollment request")))
+EMAIL_ENROLLMENT_DECISION_TITLE_PATTERN = (
+    string_concat("[%s] ", _("Your enrollment request")))
+
+VALIDATION_ERROR_USER_NOT_CONFIRMED = _(
+    "This user has not confirmed his/her email.")
```
inducer: What made you decide to define these? I'm asking because it seems easier and more direct to just keep messages where they're used. And it's easier to tell what a message says by looking at what it actually says vs. a symbolic name representing what it says.

course/enrollment.py
inducer: And it eliminates the option of matching (say) `@math.somewhere.edu` and `@cs.somewhere.edu` with just `somewhere.edu`. Regular expressions would be the real solution here...

inducer: I'm mainly skeptical because it's a change in behavior.

course/enrollment.py
dzhuang: Sorry, I think it's easier to test whether the correct messages are displayed in the page requested, in unittest. I will revert that because your opinion is more reasonable than mine.

course/enrollment.py
dzhuang: What about this? We only need to match if it endswith ".somewhere.edu" or "@somewhere.edu"

course/enrollment.py
```diff
@@ -210,10 +231,18 @@ def enroll_view(request, course_identifier):
...
+            return email.endswith(suffix)
+        else:
+            return email.endswith("@%s" % suffix) or email.endswith(".%s" % suffix)
+
     if (
             preapproval is None
             and course.enrollment_required_email_suffix
-            and not user.email.endswith(course.enrollment_required_email_suffix)):
+            and not is_email_suffix_matches(
+                user.email, course.enrollment_required_email_suffix)):
```
dzhuang: It's now checked here.

dzhuang:
Updated.

course/enrollment.py
```diff
@@ -210,10 +231,18 @@ def enroll_view(request, course_identifier):
...
+            return email.endswith(suffix)
+        else:
+            return email.endswith("@%s" % suffix) or email.endswith(".%s" % suffix)
+
     if (
             preapproval is None
             and course.enrollment_required_email_suffix
-            and not user.email.endswith(course.enrollment_required_email_suffix)):
+            and not email_suffix_matches(
+                user.email, course.enrollment_required_email_suffix)):
```
dzhuang: It's now checked here.

course/enrollment.py
```diff
@@ -210,10 +231,18 @@ def enroll_view(request, course_identifier):
                         pass
             pass
 
+    def email_suffix_matches(email, suffix):
+        # type: (Text, Text) -> bool
+        if suffix.startswith("@"):
+            return email.endswith(suffix)
+        else:
+            return email.endswith("@%s" % suffix) or email.endswith(".%s" % suffix)
```
inducer: I'd prefer this at the top level, to clarify that it's not using lexical context.

inducer: Otherwise I like this solution. :)

course/enrollment.py
inducer: Thanks for being accommodating! The tests will highlight when they're out of sync, so it should be an easy fix.

course/enrollment.py
inducer: LGTM.

inducer:
Thanks for submitting this!

course/enrollment.py
dzhuang: shall I pull them out?

inducer:
> shall I pull them out? 

Not super important, but next time you're in this bit of the code,
maybe. :)

PR #376, opened 2017-11-17
Filter endraw.

davis68:
How about this?  We filter on the line before it hits `unprocessed_block_lines`.


davis68:
Hmm... RELATE may not be happy about the indentation as it is now.  What do you think?

davis68:
Is there a unit test that covers this scenario (with embedded endraws)?

inducer:
There's still a linter issue: https://travis-ci.org/inducer/relate/jobs/303612682#L507

> Is there a unit test that covers this scenario (with embedded endraws)?

Not yet, but it'd be fantastic to have one. :)

davis68:
I've poked around with it in scaffolded isolation, but let me put a test case together before merging into the main body.

davis68:
https://github.com/inducer/relate/blob/master/tests/test_pages.py

It looks like the lion's share of existing tests test response codes rather than content rendering.  What would you prefer for testing rendering?

inducer:
> It looks like the lion's share of existing tests test response codes rather than content rendering. What would you prefer for testing rendering?

Just add another test file to test content rendering. One test is better than zero tests. :)

dzhuang:
What's the use case? Can you provide an example or documentation, perhaps I can help to write the testcases.

inducer:
Since our current YAML processing uses `{% raw %}`/`{% endraw %}` to prevent Jinja expansion in block literals, any occurrence of those inside of that block literal will cause trouble. The test would need to verify that these are now safe to use inside those block literals.

FYI, block literals are these things:

```

prompt: |

   stuff <- this is inside a block literal
```

dzhuang:
I'll have a try.

davis68:
I've had a chance to dig back through the commits causing the original problem that led to this.

The offending code was Bash, to be displayed thus:

    value=${#1}

The braces were parsed by the RELATE YAML/Jinja parser, so next I tried:

    value=${{{#}}1}
    value=${{ {# }}1}
    value=${{ '{#' }}1}

The latter works, but Andreas also suggested the expedient of using the `endraw` tags.  Then we realized that the parser was interpreting things wrongly there.

The next time it came up was when attempting to display LaTeX code:

    \newcommand{\superscript}[1] {\ensuremath{^{\textrm{#1}}}}

(You can imagine how Jinja feels about that block.  I ended up posting a screenshot instead.)  But

    {% raw %}\newcommand{\superscript}[1] {\ensuremath{^{\textrm{#1}}}}{% endraw %}

caused the sort of issue we're attempting to solve here.

davis68:
Anyway, my gumption trap on this particular PR now is that I don't know where the output is actually rendered from, nor how to catch it, in this case, and thus be able to write an actual unit test to see how this comes out.  Enlightenment appreciated.

dzhuang:
@davis68 Thanks for the details above. The integration test can be done via sandbox, not difficult, but I have time to do this until tomorow.

dzhuang:
I put the following script in the sandbox:
```yaml
type: Page
id: starting
content: |
    # Title
    {% raw %}
    {% endraw %}
```
It failed with:

>Page failed to load/validate: ScannerError: while scanning for the next token found character '%' that cannot start any token in "", line 8, column 2: {% endraw %} ^

dzhuang:
It seems the work around is to add a common indentation when replace `{% endraw %}` to `{% endraw %}\n{{ '{% endraw %}' }}\n{% raw %}`

dzhuang:
```python
                line_indent = len(LEADING_SPACES_RE.match(ln).group(1))

                if '{% endraw %}' in ln:
                    endraw_ = "{%% endraw %%}\n%s{{ '{%% endraw %%}' }}\n{%% raw %%}" % (" " * line_indent, )
                    ln = ln.replace('{% endraw %}', endraw_)
```

dzhuang:
So simply remove the `\n`

```python
if '{% endraw %}' in ln:
    endraw_ = "{% endraw %}{{ '{% endraw %}' }}{% raw %}"
    ln = ln.replace('{% endraw %}', endraw_)
```

dzhuang:
And in fact, we also need to consider other forms like `{%- endraw %}`, `{%- endraw -%}`, and ``{% endraw -%}`, so using re.sub might be better.

dzhuang:
Finally end up with:
```python
IN_BLOCK_END_RAW_RE = re.compile(r"(.*)({%-?\s*endraw\s*-?%})(.*)")
```

```python
if not ln.rstrip():
    unprocessed_block_lines.append(ln)
    i += 1
    continue

line_indent = len(LEADING_SPACES_RE.match(ln).group(1))
if line_indent <= block_start_indent:
    break
else:
    ln = IN_BLOCK_END_RAW_RE.sub(
        r"\1{% endraw %}{{ '\2' }}{% raw %}\3", ln)
    unprocessed_block_lines.append(ln.rstrip())
    i += 1
```

dzhuang:
@davis68 I've created a PR in your branch, and it passed tests. Can you review it?

inducer:
@davis68 Let me know when you consider this ready for me to look at.

davis68:
It looks great to me.  Thanks for the clarification and the test, @dzhuang!

inducer:
Thanks! @dzhuang, do you understand why the diff seems to have lots of already-merged stuff caught up in it?

inducer:
Oh well. Diffed manually, looks great. Thanks @davis68 and @dzhuang for working on this!

Issue #243, opened 2016-09-28
Need suggestions on automatically graded pages


dzhuang:
I've cooked up some customized pages with randomly selected questions from a repo data file. Currently. It works well for human graded questions.  And it only works for those kind of questions, 

What should I do if I want to create automatic graded pages, for example, a ``TextQuestion``, with randomly select questions? I can manage to generate all types of correct answers for all random selected questions in that repo file, just like how I generate the question body. But it seems currently the flow page creation process and grading process are using ``page.page_desc`` which passed validation, and which is hard coded in the yaml files. One more problem, all my question and answers are generated depending on page_context and page_data.data, it seems currently validation processes are all done without/before those context are created. 

Is it possible to extend the current process without affecting existing pages? Need your suggestions.

inducer:
Hum, I'm not sure I follow what you're saying. If you'd like to randomly pick questions from a pool, then I'd recommend you use question groups along with the `shuffle` and `max_page_count` attributes:

* https://documen.tician.de/relate/flow.html#grouping
* https://github.com/inducer/relate-sample/blob/master/flows/quiz-test.yml#L45-L49

Or am I misunderstanding what you're trying to do?

dzhuang:
I wish to do it in a different way. 

The pool is in a data file in repo. Rather than code the question and answer in the yaml, I am using some python snippet and jinja templates, which are also in repo, to generate the question body and correct answer of the question. Each time a new page is created, the ``initialize_page_data`` method will pick a data randomly from the data file, and generate the question body, along with the correct answer of the question, and store them in the ``page_data``.  The benefit of this method is that I can add new questions to the pool simply by add new data to the question data file, and if I need to change the rendering of the question body/correct answers, I can just edit the jinja template.

inducer:
I see. I think I understand better now what you're trying to do. I would however prefer to build the automatic generation process into `initialize_page_data`, to avoid the pre-generation step that you currently have to go through. Also, when I thought about the page API, I've always thought of `page_desc` as just the YAML chunk corresponding to the page, and I didn't think this would (or should) change. Instead, the `make_*` page API functions should be extended to insert data into `page_desc` on the fly. Markup expansion for the question pieces already happens under the control of that code, so in principle it should be simple to just make some more variables available. I'm thinking something like:

```yaml
type: TextQuestion # or maybe GeneratedTextQuestion?
generation: |
    answers = [{"type": float, ...}]
    markup_variables = {"x": ...}
[...]

PR #378, opened 2017-11-26
Extend tests

dzhuang:
Allow (and use) static container to test code questions, along with some refactoring of unittest methods.

This is preparing for the future `GeneratedQuestions` or `VariantGeneratingQuestion` as discussed in #243.

### DIFF
--- a/course/page/code.py
+++ b/course/page/code.py
@@ -43,6 +43,9 @@ from course.page.base import (
         get_editor_interaction_mode)
 from course.constants import flow_permission
 
+# DEBUGGING SWITCH: False for 'spawn container', True for 'a faked container'
+USE_FAKED_CONTAINER = False
+
 
 # {{{ python code question
 
@@ -95,8 +98,7 @@ def request_python_run(run_req, run_timeout, image=None):
 
     docker_timeout = 15
 
-    # DEBUGGING SWITCH: 1 for 'spawn container', 0 for 'static container'
-    if 1:
+    if not USE_FAKED_CONTAINER:
         docker_url = getattr(settings, "RELATE_DOCKER_URL",
                 "unix://var/run/docker.sock")
         docker_tls = getattr(settings, "RELATE_DOCKER_TLS_CONFIG",
--- a/course/page/code.py
+++ b/course/page/code.py
@@ -43,8 +43,10 @@ from course.page.base import (
         get_editor_interaction_mode)
 from course.constants import flow_permission
 
-# DEBUGGING SWITCH: False for 'spawn container', True for 'a faked container'
-USE_FAKED_CONTAINER = False
+# DEBUGGING SWITCH:
+# True for 'spawn containers' (normal operation)
+# False for 'just connect to localhost:RUNPY_PORT' for runpy'
+SPAWN_CONTAINERS_FOR_RUNPY = False
 
 
 # {{{ python code question
@@ -98,7 +100,7 @@ def request_python_run(run_req, run_timeout, image=None):
 
     docker_timeout = 15
 
-    if not USE_FAKED_CONTAINER:
+    if SPAWN_CONTAINERS_FOR_RUNPY:
         docker_url = getattr(settings, "RELATE_DOCKER_URL",
                 "unix://var/run/docker.sock")
         docker_tls = getattr(settings, "RELATE_DOCKER_TLS_CONFIG",
--- a/course/page/text.py
+++ b/course/page/text.py
@@ -41,6 +41,8 @@ from course.page.base import (
 import re
 import sys
 
+CORRECT_ANSWER_PATTERN = string_concat(_("A correct answer is"), ": '%s'.")  # noqa
+
 
 class TextAnswerForm(StyledForm):
     # prevents form submission with codemirror's empty textarea
@@ -961,8 +963,6 @@ class TextQuestion(TextQuestionBase, PageBaseWithValue):
     def correct_answer(self, page_context, page_data, answer_data, grade_data):
         # FIXME: Could use 'best' match to answer
 
-        CA_PATTERN = string_concat(_("A correct answer is"), ": '%s'.")  # noqa
-
         for matcher in self.matchers:
             unspec_correct_answer_text = matcher.correct_answer_text()
             if unspec_correct_answer_text is not None:
@@ -970,7 +970,7 @@ class TextQuestion(TextQuestionBase, PageBaseWithValue):
 
         assert unspec_correct_answer_text
 
-        result = CA_PATTERN % unspec_correct_answer_text
+        result = CORRECT_ANSWER_PATTERN % unspec_correct_answer_text
 
         if hasattr(self.page_desc, "answer_explanation"):
             result += markup_to_html(page_context, self.page_desc.answer_explanation)
--- a/docker-image-run-py/runpy
+++ b/docker-image-run-py/runpy
@@ -29,7 +29,21 @@ import socketserver
 import json
 import sys
 import io
-from code_runpy_backend import Struct, run_code, package_exception
+try:
+    from code_runpy_backend import Struct, run_code, package_exception
+except ImportError:
+    try:
+        # When faking a container for unittest
+        from course.page.code_runpy_backend import (
+            Struct, run_code, package_exception)
+    except ImportError:
+        # When debugging, i.e., run "python runpy" command line
+        import os
+        sys.path.insert(0, os.path.abspath(
+            os.path.join(os.path.dirname(__file__), os.pardir)))
+        from course.page.code_runpy_backend import (
+            Struct, run_code, package_exception)
+
 from http.server import BaseHTTPRequestHandler
 
 PORT = 9941
--- a/tests/base_test_mixins.py
+++ b/tests/base_test_mixins.py
@@ -778,7 +778,7 @@ class FakedRunpyContainerMixin(object):
 
         super(FakedRunpyContainerMixin, cls).setUpClass()
         cls.faked_container_patch = mock.patch(
-            "course.page.code.USE_FAKED_CONTAINER", True)
+            "course.page.code.SPAWN_CONTAINERS_FOR_RUNPY", False)
         cls.faked_container_patch.start()
 
         python_executable = os.getenv("PY_EXE")
--- a/tests/base_test_mixins.py
+++ b/tests/base_test_mixins.py
@@ -763,7 +763,7 @@ class FallBackStorageMessageTestMixin(object):
             print("\n-------no message----------")
 
 
-class FakedRunpyContainerMixin(object):
+class SubprocessRunpyContainerMixin(object):
     """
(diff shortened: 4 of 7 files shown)


tests/base_test_mixins.py
```diff
@@ -136,7 +140,63 @@
 ]
 
 
-class SuperuserCreateMixin(object):
+class ResponseContextMixin(object):
```
inducer: What is a response context? (Don't reply here, add a docstring)

tests/base_test_mixins.py
```diff
@@ -513,3 +690,41 @@ def debug_print_response_messages(self, response):
...
+    @classmethod
+    def setUpClass(cls):  # noqa
+        if six.PY2:
+            from unittest import SkipTest
+            raise SkipTest('Tests for runpy is configured for PY3 only')
+
+        super(StaticDockerContainerTestMixin, cls).setUpClass()
+        cls.static_container_patch = mock.patch(
+            "course.page.code.USE_STATIC_CONTAINER", True)
+        cls.static_container_patch.start()
```
inducer: Is there a context manager (i.e. `with xyz:`) version of this?

tests/base_test_mixins.py
```diff
@@ -513,3 +690,41 @@ def debug_print_response_messages(self, response):
...
+
+class StaticDockerContainerTestMixin(object):
+    @classmethod
+    def setUpClass(cls):  # noqa
+        if six.PY2:
+            from unittest import SkipTest
+            raise SkipTest('Tests for runpy is configured for PY3 only')
+
+        super(StaticDockerContainerTestMixin, cls).setUpClass()
+        cls.static_container_patch = mock.patch(
```
inducer: How come that's a class attribute (rather than local)?

tests/base_test_mixins.py
```diff
@@ -513,3 +690,41 @@ def debug_print_response_messages(self, response):
             print("-----------message end-------------\n")
         except KeyError:
             print("\n-------no message----------")
+
+
+class StaticDockerContainerTestMixin(object):
```
inducer: I find this name misleading, since there's no Docker involved at all. `InProcessFakeContainer`?

docker-image-run-py/runpy
```diff
@@ -29,7 +29,17 @@ import socketserver
...
+try:
+    from code_runpy_backend import Struct, run_code, package_exception
+except ImportError:
+    # When running tests
+    import os
+    sys.path.insert(0, os.path.abspath(os.path.join(os.path.dirname(__file__), os.pardir)))
+    os.environ["DJANGO_SETTINGS_MODULE"] = "relate.settings"
+    import django
+    django.setup()
+    from course.page.code_runpy_backend import Struct, run_code, package_exception
```
inducer: Can you explain why this is here? (in a comment)?

inducer: As in, why the `django.setup()` hackery? If we're running tests, I'd imagine Django is already set up?

course/page/text.py
```diff
@@ -41,6 +41,8 @@
 import re
 import sys
 
+CA_PATTERN = string_concat(_("A correct answer is"), ": '%s'.")  # noqa
```
inducer: Since it's now global (so that context is no longer immediately obvious, I'd prefer `CORRECT_ANSWER_PATTERN`.

tests/base_test_mixins.py
inducer: (I.e. I don't necessary like "static" either, as an adjective.)

tests/base_test_mixins.py
dzhuang: Yes, that's
```python
with mock.patch("course.page.code.USE_STATIC_CONTAINER", True)
    blahblah
```
However, when `setUpClass` is done, it will exit the context manager,  the constant value is reset to it's original value `False`, then all tests will be using `True`, which is not expected.

dzhuang: In fact we can also patch at the class level, by a decorator for the (mixin) class, but each time we inheretate from that mixin for a testcase, we need to add the decorator for the subclass.

dzhuang: To summarize :-) :

First,  I'm faking the server at the class level (in setUpClass() rather than setUp()), for shorter test time.  Second, `mock.patch("course.page.code.USE_STATIC_CONTAINER", True)` should be used when we are faking the runpy server, there's no need to mock `USE_STATIC_CONTAINER` locally (and it's easy to forget to add the decorator or context manager when codeing).

dzhuang: The bad side of this is we have to put altogether all tests which use faked server, excluding possible tests which use read docker container.

tests/base_test_mixins.py
dzhuang: Done.

docker-image-run-py/runpy
dzhuang: Removed.

course/page/text.py
dzhuang: Done.

tests/base_test_mixins.py
inducer: That's fine. We don't have any 'actual Docker' tests just yet, and I'm OK putting that off until we do. It's more important (and easier to debug, and not a security concern) if we do things in-process.

inducer:
Let me know when this is ready for the next round of review.

dzhuang:
It's ok for review now.

tests/test_sandbox.py
```diff
@@ -25,47 +25,125 @@
...
+value: 5\r
+prompt: |\r
+    # A half\r
+    What's a half?\r
+answers:\r
+    - type: float\r
+      value: 0.5\r
+      rtol: 1e-4\r
+    - <plain>half\r
+    - <plain>a half
```
inducer: Are these `\r`s necessary?

tests/base_test_mixins.py
```diff
@@ -513,3 +735,50 @@ def debug_print_response_messages(self, response):
...
+    def setUpClass(cls):  # noqa
+        if six.PY2:
+            from unittest import SkipTest
+            raise SkipTest("In process fake container is configured for "
+                           "PY3 only, since currently runpy docker only "
+                           "provide PY3 envrionment")
+
+        super(InProcessFakeContainerTestMixin, cls).setUpClass()
+        cls.fake_static_container_patcher = mock.patch(
+            "course.page.code.USE_STATIC_CONTAINER", True)
```
inducer: Here's another "static".

tests/base_test_mixins.py
```diff
@@ -513,3 +735,50 @@ def debug_print_response_messages(self, response):
             print("-----------message end-------------\n")
         except KeyError:
             print("\n-------no message----------")
+
+
+class InProcessFakeContainerTestMixin(object):
```
inducer: Sorry--not in-process at all. It's a subprocess the way it's currently implemented. (Btw, how come youd didn't do in-process? I.e. just a function call?)

tests/base_test_mixins.py
```diff
@@ -445,6 +596,63 @@ def assertGradeHistoryItemsCount(  # noqa
...
+            logged_in_user = None
+
+        if not target_user:
+            return logged_in_user, False
+
+        if logged_in_user != target_user:
+            self.c.force_login(target_user)
+            logged_in_user_changed = True
+
+        return logged_in_user, logged_in_user_changed
```
inducer: I suppose I'd be more comfortable with this as a context manager, to make sure cleanup actually happens.

docker-image-run-py/runpy
```diff
@@ -29,7 +29,14 @@ import socketserver
 import json
 import sys
 import io
-from code_runpy_backend import Struct, run_code, package_exception
+try:
+    from code_runpy_backend import Struct, run_code, package_exception
+except ImportError:
+    # When running tests
+    import os
+    sys.path.insert(0, os.path.abspath(os.path.join(os.path.dirname(__file__), os.pardir)))
```
inducer: Sorry, could you explain again why this is necessary? If the parent interpreter was able to import `relate` and `course`, why would the child interpreter not be able to do so?

tests/base_test_mixins.py
dzhuang: Sorry, do you mean the function `main()` in `runpy`?  My reason (might be wrong), was  `runpy` is not in a package, and it's hard to import from (though not impossible). And running a subprocess is easy to terminate in TestCase. Or is there an easier way?

dzhuang: More over, the subprocess it self can be tested (and included in coverage).

tests/test_sandbox.py
dzhuang: Seems to be. I remembered without them the tests failed with bad formatting. I need to test again to make sure.

tests/base_test_mixins.py
dzhuang: I had that idea too, since you suggested,  I'll do so :)

docker-image-run-py/runpy
dzhuang: NVM. In fact, It should be like the following, so that I can use `python runpy` in console to debug:

```python
try:
    from code_runpy_backend import Struct, run_code, package_exception
except ImportError:
    try:
        # When running tests
        from course.page.code_runpy_backend import Struct, run_code, package_exception
    except ImportError:
           # When ran by "python runpy"
           import os
           sys.path.insert(0, os.path.abspath(os.path.join(os.path.dirname(__file__), os.pardir)))
           from course.page.code_runpy_backend import Struct, run_code, package_exception
```

Am I right?

course/page/code.py
```diff
@@ -43,6 +43,9 @@
         get_editor_interaction_mode)
 from course.constants import flow_permission
 
+# DEBUGGING SWITCH: False for 'spawn container', True for 'static container'
+USE_STATIC_CONTAINER = False
```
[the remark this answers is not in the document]
dzhuang: Sorry, I copied it from your originally description at [here](https://github.com/inducer/relate/blob/master/course/page/code.py#L98) and I actually don't know if there are some kind of `static container`. I'll modify that.

tests/base_test_mixins.py
inducer: That's fair. I don't mind the subprocess. But then this should be renamed to "`SubProcess...`".

course/page/code.py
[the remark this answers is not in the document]
inducer: Oh. 😊 

In that case, a "static container" was one that i spawned manually by hand so that I could watch the output of runpy in the terminal. It's still bad terminology, but it's less bad when it's my fault, right? *cough* :)

tests/test_sandbox.py
dzhuang: You're right, those trailing string are not necessary.

dzhuang:
It's ok for next round review now.

inducer:
Thanks for putting this together!

PR #374, opened 2017-10-19
Fix error when cache is not enabled.

dzhuang:

PR #380, opened 2017-12-05
Added tests for cache, and reset objects created in setUpTestData

dzhuang:
1. Added tests for #374
1. Reset objects created in `setUpTestData` across tests, see https://docs.djangoproject.com/en/dev/topics/testing/tools/#django.test.TestCase.setUpTestData. For example, this is needed when testing `update_course`.


inducer:
Looks great, thanks!

PR #388, opened 2017-12-18
Refactor tests, and unify page_ordinal -> ordinal in urlconf

dzhuang:
`page_ordinal` and `ordinal` are unified into `ordinal` in `URLCONF` and views.


course/auth.py
```diff
@@ -270,6 +270,24 @@ def stop_impersonating(request):
         raise PermissionDenied()
 
     if not hasattr(request, "relate_impersonate_original_user"):
+        # prevent user without pperm to stop_impersonating
+        my_participations = Participation.objects.filter(
+            user=request.user,
+            status=participation_status.active)
```
inducer: So `request.user` is the impersonated user, which means that getting out of impersonating someone now would require a logout?

course/grading.py
```diff
@@ -88,7 +88,7 @@ def get_prev_visit_grades(
 
 
 @course_view
-def get_prev_grades_dropdown_content(pctx, flow_session_id, page_ordinal):
+def get_prev_grades_dropdown_content(pctx, flow_session_id, ordinal):
```
inducer: I think I would've preferred standardizing the other way. `ordinal` is less descriptive than `page_ordinal`.

course/auth.py
dzhuang: No, it was preventing who didn't currently impersonating and who don't have impersonating pperm to do `stop_impersonating`. You can see from the testcase [here](https://github.com/inducer/relate/pull/388/files#diff-b4c9abe346f4b3686dd3e1025a50aaa7R116)

course/grading.py
dzhuang: Ok. Is it necessary to standardize that name in model?  (needs migrations).

dzhuang: Updated, wth `FlowPageData.ordinal` renamed to `page_ordinal`.

inducer: :/

course/auth.py
dzhuang: Without this, student user can view the `impersonate` page. This patch redirects them to 403.

dzhuang: I know that requires extra db hits when stop impersonating, but I just want them to know there are permissions.

inducer:
Thanks!

course/auth.py
dzhuang: > This patch redirects them to 403.

The tests is (here)[https://github.com/inducer/relate/pull/388/files#diff-b4c9abe346f4b3686dd3e1025a50aaa7R81]. Without this, the `status_code` is 200(302).

inducer: Sure. It doesn't hurt. Thanks for working on this!

PR #390, opened 2017-12-21
Used tempfile.mkdtemp for GIT_ROOT in tests

dzhuang:
Reduce the need to delete the git folders created and skip tests when course creation failed.


dzhuang:
It seems I'll have to add a test with validation failure so as to get passed :)

inducer:
Thanks!

PR #391, opened 2017-12-22
Fixed bug (see #391) and sped up course creation for tests.

dzhuang:
* Bug: [this line](https://github.com/inducer/relate/blob/master/course/auth.py#L109) happened to get passed tests (now it won't).
* Use cache to memorize the git repo path of previously created course, to speed up tests.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -9,6 +9,7 @@ build: off
 
 cache:
   - '%LOCALAPPDATA%\pip\Cache'
+  - memcached -> appveyor.yml
 
 environment:
   global:
@@ -39,6 +40,10 @@ install:
   # done from inside the powershell script as it would require to restart
   # the parent CMD process).
   - "SET PATH=%PYTHON%;%PYTHON%\\Scripts;%PATH%"
+
+  - appveyor DownloadFile "http://s3.amazonaws.com/downloads.northscale.com/memcached-win64-1.4.4-14.zip" -FileName memcached.zip
+  - 7z x memcached.zip -y
+  - ps: $Memcached = Start-Process memcached\memcached.exe -PassThru
 
   - ECHO "Filesystem root:"
   - ps: "ls \"C:/\""
--- a/course/auth.py
+++ b/course/auth.py
@@ -106,7 +106,6 @@ def get_impersonable_user_qset(impersonator):
             if perm == pperm.impersonate_role]
 
         q = (Participation.objects
-             .exclude(pk=impersonator.pk)
              .filter(course=part.course,
                      status=participation_status.active,
                      roles__identifier__in=impersonable_roles)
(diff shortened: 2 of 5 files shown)


inducer:
> Bug: this line happened to get passed tests (now it won't).

Can you explain?

dzhuang:
> Can you explain?

The involved [test](https://github.com/inducer/relate/blob/master/tests/test_auth.py#L150-L153) is:

```python
    def test_impersonate_by_instructor(self):
        user = self.instructor_participation.user
        impersonatable = get_impersonable_user_qset(user)
        self.assertEqual(impersonatable.count(), 2)
```
The expected impersonatable when instructor is impersonating is a "ta", and a "student" (count for 2). 

For test-course, the superuser (with user pk=1) is participation pk=1(because by post_create_new_course, the superuser is always the first participation) , instructor (with user pk=2) is participation=2, the `exlclude` operation below tries to exclude the instructor's participation(pk=2), while 
the wrong code excluded his user.pk (which is also 2). So it passed test. [src](https://github.com/inducer/relate/blob/master/course/auth.py#L108-L113)
```python
        q = (Participation.objects
             .exclude(pk=impersonator.pk)
             .filter(course=part.course,
                     status=participation_status.active,
                     roles__identifier__in=impersonable_roles)
             .select_related("user"))
```

Now with the lastest patch, if the course is created from db in a testcase, instructor now has user.pk=2, with his participation pk=1 (superuser is no longer the first participation), the above script excludes participation pk=2 (which is ta's participation), the result is wrong.

Actually, that exclusion is totally necessary because the impersonator himself had been excluded in [pperm filter stage(https://github.com/inducer/relate/blob/master/course/auth.py#L103-L106), i.e.,

```python
        impersonable_roles = [
            argument
            for perm, argument in part.permissions()
            if perm == pperm.impersonate_role]
```

dzhuang:
The build time (for PY3.5, which last the longest) reduced about 20 seconds (from around 70s to 35s) both on Travis-ci and Appvayor. For my local tests, that reduce significantly, sometimes I need 10 seconds to fetch from the relate-sample repo.

inducer:
> Because one participation cannot impersonate another one of the same role.

By default, that's not true, but I could see this being configured for some courses, e.g. where one instructor might impersonate another.

inducer:
So, considering the last comment, the impact of this is that, in a non-default permissions situation (say instructor may impersonate another instructor), someone might be able to impersonate themselves, right? Is that bad? Not sure it matters.

dzhuang:
So the situation is there're two instuctors in a same course, and with a non-default configuration on pperm? I don't think it's a big deal, currently.

inducer:
Ok, I agree. (And I also agree that the previous code is incorrect.)

inducer:
Thanks for working on this!

dzhuang:
Actually (I remembered) that bug was produced by my pr, so I'm responsible for that :)

Issue #383, opened 2017-12-15
FeatureRequest: Allow instructor to determine the active language used in a course


dzhuang:
And this will override user preference proposed in #382

dzhuang:
Closed via #392.

PR #392, opened 2017-12-23
Added course specific language

dzhuang:
Allow course specific language #383


dzhuang:
The naming and descrpitions need to be polished.

course/utils.py
```diff
@@ -659,7 +660,19 @@ def has_permission(self, perm, argument=None):
...
+            if action == "activate":
+                translation.activate(self.course.force_lang)
+            else:
+                translation.deactivate()
+
     def __enter__(self):
+        self._set_course_lang(action="activate")
         return self
 
     def __exit__(self, exc_type, exc_val, exc_tb):
```
inducer: How come this makes no effort to restore the prior state on `__exit__`?

course/migrations/0110_add_force_lang_field_to_course.py
```diff
@@ -0,0 +1,20 @@
...
+        ('course', '0109_add_manage_authentication_tokens_permssion'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='course',
+            name='force_lang',
+            field=models.CharField(blank=True, default=None, help_text="Which language name is forced to be used for this course. For example, 'en', 'de', 'zh-hans'. If not set, displayed language will be determined by user preference", max_length=200, null=True, verbose_name='Course language forcibly used'),
+        ),
+    ]
```
inducer: I'm not totally opposed to this, but I'd like to understand the use case a bit better. What's the reason to not just trust the browser when it reports language preference to the server?

dzhuang: For example, in some universities in China, there are some courses which are required to be taught in English ( including course materials), other than the native language.  It will be weird to have context (like date, menu and literals like "choices", "correct answers") in Chinese while the contents are in English.

course/utils.py
dzhuang: Thanks, I'll try that.

dzhuang: Sorry, I did do the restoring but I didn't know why that line just disappear...

dzhuang: Updated.

course/migrations/0110_add_force_lang_field_to_course.py
inducer: I see. Fair enough. :)

course/utils.py
```diff
@@ -659,10 +660,25 @@ def has_permission(self, perm, argument=None):
...
             return (perm, argument) in self.permissions()
 
+    def _set_course_lang(self, action):
+        # type: (Text) -> None
+        from django.conf import settings
+        if not getattr(settings, "RELATE_ENABLE_COURSE_SPECIFIC_LANG", False):
+            return
+        if self.course.force_lang is not None:
+            if action == "activate":
+                self.old_language = translation.get_language()
```
inducer: This should assert that `old_language` had not previously been set, to avoid incorrect behavior if this context manager is used in a nested manner.

course/models.py
```diff
@@ -192,6 +192,14 @@ class Course(models.Model):
             "notifications about the course."),
             verbose_name=_('Notify email'))
 
+    force_lang = models.CharField(max_length=200, blank=True, null=True,
+            default=None,
+            help_text=_(
+                "Which language is forced to be used for this course. "
+                "If not set, displayed language will be determined by "
+                "user browser preference"),
```
inducer: * The help text should specify how that language is encoded. (likely by referencing Django's language codes)
* Whole sentences get periods at the end. :)

course/utils.py
```diff
@@ -1099,4 +1115,16 @@ def will_use_masked_profile_for_email(recipient_email):
             return True
     return False
 
+
+def get_course_specific_langs_choices():
```
inducer: `langs` -> `languages`

course/utils.py
```diff
@@ -1099,4 +1115,16 @@ def will_use_masked_profile_for_email(recipient_email):
...
+    # type: () -> Tuple[Tuple[str, Any], ...]
+    from django.conf import settings
+    return (("", _("(None)")), ) + tuple((k, _(v)) for (k, v) in settings.LANGUAGES)
+
+
+def get_available_languages():
+    # type: () -> List[Text]
+    from django.conf import settings
+    return [lang[0] for lang in settings.LANGUAGES]
```
inducer: Would prefer `[lang_code for lang_code, descr in settings.LANGUAGES]`.

course/versioning.py
```diff
@@ -183,11 +185,18 @@ class Meta:
...
             "from_email",
             "notify_email",
+            "force_lang",
             )
         widgets = {
                 "start_date": DateTimePicker(options={"format": "YYYY-MM-DD"}),
                 "end_date": DateTimePicker(options={"format": "YYYY-MM-DD"}),
                 }
+        from django.conf import settings
+        if getattr(settings, "RELATE_ENABLE_COURSE_SPECIFIC_LANG", False):
```
inducer: Do we need `RELATE_ENABLE_COURSE_SPECIFIC_LANG`? It just adds code complexity, and instructors that don't want it can just leave it alone.

course/versioning.py
```diff
@@ -204,6 +213,23 @@ def clean_git_source(self):
 
         return self.cleaned_data["git_source"]
 
+    def clean_force_lang(self):
+        force_lang = self.cleaned_data["force_lang"]
+
+        if not force_lang:
```
inducer: Code duplicated with below. Maybe introduce a widget to share it?

dzhuang: That's a good idea.

course/models.py
dzhuang: > The help text should specify how that language is encoded. (likely by referencing Django's language codes)

I'm wondering is that helpful? Because the user (instructor) just see the (translated) description, there's no access to the language code in the form.

inducer: If someone is looking at this in the admin, possibly.

dzhuang: Ok. However, there's another case (for me). I used `zh-cn` instead of `zh-hans`  as `Simplified Chinese` through customized settings.  If an instructor refer to the Django document, he will be misled to a wrong languague code. So a better solution might be concatenate the language_code to the language description?

course/utils.py
dzhuang: Do you mean
```python
if action == "activate" and self.old_language is not None?
```

course/models.py
dzhuang: I just realize that help_text is assuming the settings has `USE_I18n` enabled. I.e., that description can not cover when it is not enabled.

course/utils.py
inducer: No. I meant that

* if we don't expect nested use: this should crash if `old_language` is already set (requires that `old_language` be initialized in the constructor),
* if we do expect nested use: use a list/stack of `old_languages` instead.

IOW, have a definite answer for what should happen if use of this context mgr is nested.

course/models.py
inducer: Yeah, that sounds reasonable.

course/versioning.py
dzhuang: It seems that should be done by a `validators` param in the the field definition, not widget.

course/utils.py
dzhuang: Then it can be view as a good side effect to check whether the `course_view` is nestedly (duplicated) used, for developers, am I right? Then why not add another property such as `_in_context_manager` to indicate that?

course/versioning.py
dzhuang: Ok, that really simplified the jobs.

course/utils.py
dzhuang: Ok.

course/utils.py
dzhuang: Done.

course/versioning.py
dzhuang: Done.

dzhuang:
Updated.

course/utils.py
inducer: That's fine, too.

inducer:
LGTM aside from the nesting check.

course/utils.py
```diff
@@ -659,10 +664,28 @@ def has_permission(self, perm, argument=None):
...
+                    if self.old_language is not None:
+                        translation.activate(self.old_language)
+
     def __enter__(self):
+        if self._is_in_context_manager:
+            raise RuntimeError(
+                "Nested use of 'course_view' as context manager "
+                "is not allowed.")
+        self._is_in_context_manager = True
+        self._set_course_lang(action="activate")
```
dzhuang: What about this? @inducer

dzhuang: Or do you mean I need to add test on that?

inducer: No, this is good.

inducer:
Thanks!

PR #397, opened 2018-01-04
Update python-memcached

orsenthil:
Fixing this failure

```
Collecting python-memcached==1.58 (from -r requirements.txt (line 81))
  Using cached python-memcached-1.58.tar.gz
    Complete output from command python setup.py egg_info:
      File "<string>", line 1, in <module>
      File "/private/var/folders/yx/tn9fxy1s4qz1vf2cbb5fwlw00000gn/T/pip-build-l3yvlcg7/python-memcached/setup.py", line 8, in <module>
        version=get_module_constant('memcache', '__version__'),
      File "/Users/senthil/py/virtualenvironments/relateenv/lib/python3.6/site-packages/setuptools/depends.py", line 164, in get_module_constant
        return extract_constant(code, symbol, default)
      File "/Users/senthil/py/virtualenvironments/relateenv/lib/python3.6/site-packages/setuptools/depends.py", line 195, in extract_constant
        const = code.co_consts[arg]
    IndexError: tuple index out of range

    ----------------------------------------
Command "python setup.py egg_info" failed with error code 1 in /private/var/folders/yx/tn9fxy1s4qz1vf2cbb5fwlw00000gn/T/pip-build-l3yvlcg7/python-memcached/
```

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -78,7 +78,7 @@ git+https://github.com/knaperek/djangosaml2.git
 # Note: djangosaml2 uses a non-Py3 version of python-memcached directly.  #
 # Override that. Also: Don't be tempted to use this for Django caching--it's
 # broken.
-python-memcached==1.58
+python-memcached==1.59
 
 # }}}
 


inducer:
Hm. This looks like an issue in setuptools more than in python-memcached. I'm also surprised we're not seeing this failure in our [builds](https://travis-ci.org/inducer/relate/builds). Also, the failing llne of python-memcached's [`setup.py`](https://github.com/linsomniac/python-memcached/blob/master/setup.py) seems unchanged.

Don't get me wrong, I'm not in principle against upgrading python-memcached, but I'm not sure that this update will have an impact on the failure you cite.

orsenthil:
@inducer  - I was able to try `relate` locally after making this change. It seems to me that is a problem with python-memcached-1.58

orsenthil:
Okay, it is an issue with setuptools - https://github.com/linsomniac/python-memcached/issues/103#issuecomment-269938185 and https://github.com/pypa/setuptools/issues/866 
I was trying it with Python3.6 to land on this issue.

@inducer  - take your call on upgrading python-memcached or dropping this PR.

inducer:
> I'm also surprised we're not seeing this failure in our builds.

We weren't because we pin the version of setuptools.

inducer:
Upgrading seems harmless, so why not--even though I still don't see the connection between this issue and 1.59.

Issue #355, opened 2017-09-04
Theme for relate


samtux:
I share a theme for this great software. 

https://github.com/samtux/relate-theme

inducer:
Thanks for sharing this! This looks really nice. I would love to integrate this, if there were a way to have both the "classic" and this (and potentially more) themes.

samtux:
I'm glad you support the topic and the possibility of adding others. In the future I would like to support with other Bootstrap themes.  

I don't know what would be the best recommendation to support several themes in Relate, in this case I had to edit several Django templates, and the style sheets and libraries were stored in /static/theme/.

dzhuang:
https://docs.djangoproject.com/en/dev/howto/overriding-templates/

We can allow user to override the template (by adding template dir or insert custome apps before `accounts` and `course` in `INSTALLED_APPS` in settings.py, or futher, create some common template that allow "classic" and other themes to extend.

dzhuang:
As a note from #398, if you need to override just the `style.css` file, and leaving other built-in staticfiles untouched, you can just prepend the staticfile dir to the RELATE default `STATICFILES_DIRS`, by overriding `STATICFILES_DIRS` in local_settings.py. The structure should look like:

```none
project/
        relate/
               static/
                       css/
                            style.css
                       images/
                            busy.gif
                       js/
                            dataTables.relate.sort.plugin.js
        my_static/
                       css/
                            style.css
```

dzhuang:
Thus, switching between different themes can be done by putting different staticfiles folder at the begining of `STATICFILES_DIRS`, along with different `TEMPLATE` `DIRS` settings (proposed in #398).

PR #398, opened 2018-01-04
Use settings to control relate instance brand name, allow template overriding.

dzhuang:


local_settings.example.py
```diff
@@ -257,6 +265,36 @@
...
+#                      os.path.join(os.path.dirname(__file__), "my_other_templates")]
+
+# Uncomment the following to enable staticfile overriding, It should be configured
+# as a list/tuple of path(s)
+
+# MY_STATICFILES_DIRS = [os.path.join(os.path.dirname(__file__), "staticfiles"),]
+
+# }}}
 
 # {{{ docker
```
inducer: Not a fan of `MY_` prefixes--I'd prefer if these said `RELATE_` instead.

inducer:
LGTM except for the settings rename.

local_settings.example.py
```diff
@@ -257,6 +265,36 @@
...
+#
+
+# import os.path
+# MY_TEMPLATES_DIRS = [os.path.join(os.path.dirname(__file__), "my_templates"),
+#                      os.path.join(os.path.dirname(__file__), "my_other_templates")]
+
+# Uncomment the following to enable staticfile overriding, It should be configured
+# as a list/tuple of path(s)
+
+# MY_STATICFILES_DIRS = [os.path.join(os.path.dirname(__file__), "staticfiles"),]
```
inducer: Doesn't Django already have a way of overriding these? If so, is it really necessary to add these?

dzhuang: Yes, the overriding of staticfiles works like that of Django templates. These are shortcut for overriding, it make it easier especially for template (there will be less work when settings.py is updated), and I agree it's less necessary for staticfiles. I added that mainly because #355 is adding staticfiles to the project for the new theme. I'll remove it.

dzhuang: For staticfiles, perhaps it's only necessary when more work can be done on proposals like #355.

local_settings.example.py
dzhuang: Updated to `RELATE_OVERRIDE_`

local_settings.example.py
dzhuang: Removed.

relate/settings.py
```diff
@@ -255,7 +261,13 @@
 
 # {{{ celery config
 
-BROKER_URL = 'django://'
+RELATE_SITE_NAME = gettext_noop("RELATE")
+RELATE_CUTOMIZED_SITE_NAME = local_settings.get("RELATE_CUTOMIZED_SITE_NAME")
+if RELATE_CUTOMIZED_SITE_NAME is not None and RELATE_CUTOMIZED_SITE_NAME.strip():
+    RELATE_SITE_NAME = RELATE_CUTOMIZED_SITE_NAME
```
inducer: Could you move this outside of the "celery config" section?

dzhuang: Ops.., my bad. Updated.

dzhuang:
@inducer The PR is ready for review.

BTW, about the work flow. Do you feel reviewing squashed commits for PR a problem? As you can see, I wish to reduce the number of commits (PRs submitted by me) in your master branch, but that might bring extra work for you to do the review, especially when a PR were modified serveral times. If that's true, I can submit changes on the PR (without squashing), and you can tell me to squash the commits if the PR is approved.

inducer:
> Do you feel reviewing squashed commits for PR a problem?

I certainly don't like it, because it becomes difficult to find comments I might have left on a prior version.

> As you can see, I wish to reduce the number of commits

I don't necessarily share that objective.

Squash after approval also doesn't seem necessary. If needed, github can "squash and merge" upon merge. But even so, I'm not likely to use that...

dzhuang:
Ok, I'll adjust my work flow. Thanks.

relate/utils.py
```diff
@@ -163,6 +165,21 @@ def __call__(self, request):
...
+    # type: () -> Text
+    from django.conf import settings
+    return getattr(settings, "RELATE_SITE_NAME", "RELATE")
+
+
+def render_email_template(template_name, context=None, request=None, using=None):
+    # type: (Text, Optional[Dict], Optional[HttpRequest], Optional[bool]) -> Text
+    if context is None:
+        context = {}
+    context.update({"RELATE": _(get_site_name())})
```
inducer: I kind of feel that `{{ site_name }}` or `{{ relate_site_name }}` would be more self-explanatory.

course/templates/course/sign-in-email.txt
```diff
@@ -8,4 +8,4 @@ You have received this email because someone (maybe you) entered your email addr
 
 If this was not you, it is safe to disregard this email.
 {% endblocktrans %}
-{% blocktrans %}- RELATE staff {% endblocktrans %}
+- {{ RELATE }} {% trans "staff" %}
```
inducer: This seems to be hardcoding the phrasing "XYZ staff", which may or may not work in some languages.

relate/utils.py
inducer: (Plus, the upper-case variable name is a bit weird.)

dzhuang: I agree. Though my current code is to reduce the work of changing the i18n literals, I'll change that.

course/templates/course/sign-in-email.txt
dzhuang: Ok.

relate/utils.py
dzhuang: Will change to `relate_site_name`.

dzhuang:
Updated. P.S., It seems codecov had set wrong reference to compare the coverage data between commits.

course/templates/course/home.html
```diff
@@ -2,16 +2,18 @@
 {% load i18n %}
 
 {% block content %}
+  {% block cover %}
   <div class="jumbotron">
-    {% trans "RELATE" as RELATE %}
-    <h1>{% blocktrans %} Welcome to {{ RELATE }} {% endblocktrans %}</h1>
+    <h1>{% blocktrans %} Welcome to {{ relate_site_name }} {% endblocktrans %}</h1>
```
inducer: Shouldn't that use a `{% with` block to feed in the site name? Otherwise all the translation literals need to be changed.

course/templates/course/home.html
```diff
@@ -2,16 +2,18 @@
...
     <p>{% trans "RELATE is an Environment for Learning And TEaching" %}</p>
```
inducer: Did you want to change this, too?

dzhuang: I don't think so. Because that's an abbr. If people want to replace that, they should do by template overriding.

inducer: Fair point.

course/templates/course/home.html
dzhuang: You're right. How come I didn't think of that...

dzhuang: Updated.

course/templates/course/analytics-flow.html
```diff
@@ -1,8 +1,9 @@
 {% extends "course/course-base.html" %}
 {% load i18n %}
+{% trans relate_site_name as RELATE %}
 
 {% block title %}
-  {% trans "Analytics" %} - {% trans "RELATE" %}
+  {% trans "Analytics" %} - {{ RELATE }}
```
inducer: Shouldn't this be `relate_site_name`?

course/templates/course/analytics-flows.html
```diff
@@ -1,8 +1,9 @@
 {% extends "course/course-base.html" %}
 {% load i18n %}
+{% trans relate_site_name as RELATE %}
 
 {% block title %}
-  {% trans "Analytics" %} - {% trans "RELATE" %}
+  {% trans "Analytics" %} - {{ RELATE }}
```
inducer: Shouldn't this be `relate_site_name`?

course/templates/course/analytics-page.html
```diff
@@ -1,8 +1,9 @@
 {% extends "course/course-base.html" %}
 {% load i18n %}
+{% trans relate_site_name as RELATE %}
 
 {% block title %}
-  {% trans "Analytics" %} - {% trans "RELATE" %}
+  {% trans "Analytics" %} - {{ RELATE }}
```
inducer: Shouldn't this be `relate_site_name`?

course/templates/course/calendar.html
```diff
@@ -1,11 +1,12 @@
 {% extends "course/course-base.html" %}
 {% load i18n %}
+{% trans relate_site_name as RELATE %}
 
 {% load static %}
 
 {% block title %}
   {{ course.number}}
-  {% trans "Calendar" %} - {% trans "RELATE" %}
+  {% trans "Calendar" %} - {{ RELATE }}
```
inducer: Shouldn't this be `relate_site_name`?

relate/templates/base.html
```diff
@@ -1,15 +1,17 @@
 <!DOCTYPE html>
 {% load i18n %}
 {% load static %}
+{% trans relate_site_name as RELATE %}
```
inducer: Why? I'm not that happy with `RELATE` as a variable name.

course/templates/course/sign-in-email.txt
```diff
@@ -8,4 +8,4 @@ You have received this email because someone (maybe you) entered your email addr
 
 If this was not you, it is safe to disregard this email.
 {% endblocktrans %}
-{% blocktrans %}- RELATE staff {% endblocktrans %}
+- {% blocktrans %}{{ RELATE }} staff{% endblocktrans %}
```
inducer: Does this inherit identifiers?

course/templates/course/submit-notify.txt
```diff
@@ -1,8 +1,8 @@
-{% load i18n %}{% blocktrans with course_identifier=course.identifier %}Dear course staff of {{ course_identifier }},{% endblocktrans %}
+{% load i18n %}{% trans relate_site_name as RELATE %}{% blocktrans with course_identifier=course.identifier %}Dear course staff of {{ course_identifier }},{% endblocktrans %}
  
 {% if flow_session.participation and not use_masked_profile %}{% blocktrans with participant=flow_session.participation.user %}Participant '{{ participant }}'{% endblocktrans %}{% else %}{% trans "A participant" %}{% endif %}{% blocktrans with flow_id=flow_session.flow_id course_identifier=course.identifier %} has just submitted his/her work on '{{ flow_id }}'.
 
 Click here to review it:
 {{ review_uri }}
 {% endblocktrans %}
-{% blocktrans %}- RELATE staff {% endblocktrans %}
+- {% blocktrans %}{{ RELATE }} staff{% endblocktrans %}
```
inducer: Does this inherit identifiers?

relate/templates/base.html
dzhuang: sorry, it seems I misunderstood you previous comments. Can you revert this commit and modify the previous one? thanks.

course/templates/course/submit-notify.txt
dzhuang: `{% trans relate_site_name as RELATE %}` at the beginning has make 'RELATE' a varible in the template.

relate/templates/base.html
dzhuang: So, let me guess what you meant.

1. Use `relate_site_name` when the literals IS 'relate-site_name' itself, no need to change to `RELATE`.

1. Avoid translation literals which *contain* `RELATE` in template.

1. For Python literals, avoid use `RELATE` as variable name (if we did that before, correct it).

Am I right?

inducer: Yep, that sounds right. Sorry for not having been clear and having caused you extra work.

inducer:
LGTM. Thanks!

relate/templates/base.html
dzhuang: NVM.

dzhuang:
Thank you for reviewing the code, I know it's difficult for large PRs.

PR #401, opened 2018-01-09
Ensure uniqueness of user email.

dzhuang:
Without this, staff can save multiple users with the same email in admin. Then reset password through email will result in 500 for those users.

### DIFF
--- a/accounts/models.py
+++ b/accounts/models.py
@@ -236,7 +236,7 @@ class User(AbstractBaseUser, PermissionsMixin):
 
         # This is for backward compatibility.
         # Because user instances are frequently updated when auth_login,
-        # reset_password. Without this, all user won't be able to login.
+        # reset_password. Without this, no user will be able to login.
         if ((update_fields is not None and "email" in update_fields)
                 or self.pk is None):
             self.clean()
--- a/accounts/models.py
+++ b/accounts/models.py
@@ -220,7 +220,30 @@ class User(AbstractBaseUser, PermissionsMixin):
 
         return _("user")
 
+    def clean(self):
+        super(User, self).clean()
+        qset = self.__class__.objects.filter(email__iexact=self.email)
+        if self.pk is not None:
+            # In case editing an existing user object
+            qset = qset.exclude(pk=self.pk)
+        if qset.exists():
+            from django.core.exceptions import ValidationError
+            raise ValidationError(
+                {"email": _("That email address is already in use.")})
+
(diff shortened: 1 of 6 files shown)


inducer:
I think that this is *in principle* a good idea. One thing that's bothering me a bit about this is that it partially conveys a false sense of security, because email addresses are case-insensitive (as specified in the RFCs).

There are [*some* ways](https://stackoverflow.com/questions/7773341/case-insensitive-unique-model-fields-in-django), to deal with that, though none look fantastic.

In summary, I'm OK merging this, but I figured I would give you a chance to respond to this concern before I do.

dzhuang:
Thanks for the information, I didn't realize that before. Can we work round it by modify the `save` method?

inducer:
Or a validator?

dzhuang:
> Or a validator?

That sounds reasonable. I'll check that.

dzhuang:
Done by add override clean in `accounts.model.User`.  This PR is ready for review. @inducer

course/auth.py
```diff
@@ -583,80 +585,95 @@ def reset_password(request, field="email"):
...
+                # This is for backward compatibility.
                 messages.add_message(request, messages.ERROR,
-                        _("That %(field)s doesn't have an "
-                            "associated user account. Are you "
-                            "sure you've registered?")
-                        % {"field": FIELD_DICT[field]})
+                        _("Failed to send an email: multiple users were "
+                          "unexpectedly using that same "
+                          "email address. Please "
+                          "contact site staffs."))
```
inducer: "staff" is already plural.

accounts/models.py
```diff
@@ -220,7 +220,30 @@ def get_email_appellation(self):
...
+            from django.core.exceptions import ValidationError
+            raise ValidationError(
+                {"email": _("That email address is already in use.")})
+
     def save(self, *args, **kwargs):
+        update_fields = kwargs.get("update_fields")
+
+        # This is for backward compatibility.
+        # Because user instances are frequently updated when auth_login,
+        # reset_password. Without this, all user won't be able to login.
```
inducer: Phrasing: "all user won't be able to login." -> one of:

* no user will be able to log in
* some users will be unable to log in.

course/auth.py
dzhuang: Thanks !

dzhuang: Done.

inducer:
Thanks!

PR #402, opened 2018-01-09
Ensure I18N literals were marked correctly via CI.

dzhuang:
If I18N literals marked incorrect, CI tests will fail.


inducer:
Thanks!

PR #405, opened 2018-01-10
Test settings checks item-wise.

dzhuang:
Filter out check messages from other settings items.  This is helpful for testing customized local settings.


inducer:
LGTM. Thanks!

PR #407, opened 2018-01-11
Fix #406 with tests.

### LINKED ISSUE #406: Bug: institutional_id can't be empty when saving profile form
It raises "field is required" or "input not matched". Why that bug had not been reported before?

dzhuang:


course/auth.py
```diff
@@ -911,7 +907,9 @@ def __init__(self, *args, **kwargs):
                     "<b>Once %(submitted_or_verified)s, it cannot be "
                     "changed</b>.")
                 % {"submitted_or_verified":
-                    EDITABLE_INST_ID_BEFORE_VERIFICATION
+                    getattr(settings,
+                            "RELATE_EDITABLE_INST_ID_BEFORE_VERIFICATION",
+                            True)
```
inducer: Could you make a function to get that setting? Otherwise the default value is repeated, which could lead to inconsistencies.

course/auth.py
```diff
@@ -937,6 +935,28 @@ def adjust_layout(is_inst_id_locked):
...
+            may_not_change_field += ["first_name", "last_name"]
+
+        if self.is_inst_id_locked:
+            may_not_change_field.append("institutional_id")
+
+        changed_data = []
+        for name in self.Meta.fields:
+            if name in data and name not in may_not_change_field:
+                changed_data.append(name)
+        return changed_data
```
inducer: Why not just fail validation if the data is changed (rather than pretend that the data didn't change)?

course/auth.py
dzhuang: That's reasonable. The realization before is using a constant, that brought trouble for tests (because overiding settings won't work for testing different configurations), but a function will be ok. Will update.

course/auth.py
dzhuang: That's a good point. What I did was ignore the change, and ignore saving of those items. Will update.

dzhuang: I think we should raise those errors in `clean_` method.

dzhuang: From the [doc](https://docs.djangoproject.com/en/dev/ref/forms/fields/#disabled)
> The disabled boolean argument, when set to True, disables a form field using the disabled HTML attribute so that it won’t be editable by users. Even if a user tampers with the field’s value submitted to the server, it will be ignored in favor of the value from the form’s initial data.

It seems we don't need to raise those errors. And, as you can see, those code were covered by tests, it seens the `changed_data` was generated before form data validation.

dzhuang: From the [doc](https://docs.djangoproject.com/en/2.0/ref/forms/api/#django.forms.Form.changed_data):
>The `changed_data` attribute returns a list of the names of the fields whose values in the form’s bound data (usually `request.POST`) differ from what was provided in `initial`. It returns an empty list if no data differs.

The diff is between POST data, rather than cleaned_data. I overrided `changed_data` so that `has_changed` can be used to determine whether we need to do a DB save, and send what messages to user. If we want to diff `cleaned_data` with initial instance, we need to implement some new method.

dzhuang: I just found out current implementation with old version of Django. Since 1.9 we an directly disable a field rather than disable its widget. Then the modification is rather simple.

dzhuang: Updated.

course/auth.py
dzhuang: Updated.

course/auth.py
```diff
@@ -1005,34 +987,45 @@ def is_inst_id_locked(user):
...
-                    is_inst_id_locked=is_inst_id_locked(request.user),
-            )
+        request.user.refresh_from_db()
+        user_form = UserForm(
+            instance=request.user,
+            is_inst_id_locked=is_inst_id_locked(request.user),
+        )
+    if is_inst_id_locked(request.user):
+        user_form.fields['institutional_id'].disabled = True
+        user_form.fields['institutional_id_confirm'].disabled = True
```
inducer: How come this is not in the `UserForm` constructor?

dzhuang: I also found the code was in a mess (composed by me). I'm rewriting this part.

course/auth.py
inducer: > I overrided changed_data so that has_changed can be used to determine whether we need to do a DB save

I still don't see how we're allowed to redefine the semantics of (and thus override) `changed_data` though--it's an interface defined by Django.

dzhuang: You are right, my new idea is to remove that, by removing the checkbox field, and use a checkbox element instead.

dzhuang: update.

course/auth.py
dzhuang: update.

inducer:
LGTM, thx!

dzhuang:
Thanks!

PR #411, opened 2018-01-15
Purge page views

inducer:
My database has grown quite large. (tens of gigabytes) This is an attempt to cut down on some of that.

@dzhuang Could you take a look?


dzhuang:
Is that attempting to purge page visit data which have no answer submission? I'm ok with that. I'll try to add tests for that. 

What asserssions were supposed to be made? 
1. Permssions
1. Not effecting other course's data if not superuser.

inducer:
> Is that attempting to purge page visit data which have no answer submission?

Yeah. The idea is that while I would like to keep any answers that students have submitted basically indefinitely, I feel that records of when what student visited what page lose their value maybe a semester after the course ended. (But they do have value to me initially--I've used them in the past to support academic integrity cases.)

To be precise: There may be answer submissions for the same pages somewhere in the database. Those should never be deleted, so a good test would be maybe setting up a basic visit instance and then making sure that their count hasn't changed.

inducer:
> What asserssions were supposed to be made?

This is intended as a combined administrative/self-service function, with the main purpose being that the site admin would email course owners about this data being deleted, and after a week going ahead and doing so. Alternatively, the admin could ask course owners to do so themselves. The functionality should support both models.

inducer:
How well can we test things that use celery from our tests?

inducer:
> I'll try to add tests for that.

Awesome, thanks!

dzhuang:
> How well can we test things that use celery from our tests?

There's some examples by quick googling, like this [project](https://github.com/pmclanahan/django-celery-email/tree/master/tests).

dzhuang:
For setting up a test instance for this purpose, I currently think of introducing FactoryBoy which can quickly setup numerous visit entries in one go. It should not be difficult.

dzhuang:
However, I need two/three days to deal with grading of my courses before I can have time to handle this.

dzhuang:
>But they do have value to me initially--I've used them in the past to support academic integrity cases.

In fact, I'm of interest on that. Is there any published paper I can reference? Thanks. (And, there might be some possiblities to make some cross-cultural studies?)

inducer:
> In fact, I'm of interest on that. Is there any published paper I can reference?

Not that I know of.

> (And, there might be some possiblities to make some cross-cultural studies?)

I assume that while I have access to that data, I'm not allowed to use it for anything but class-related purposes, at least until get approval from our internal review board.

dzhuang:
> I assume that while I have access to that data, I'm not allowed to use it for anything but class-related purposes, at least until I get approval from our internal review board.

Got it. But that is required even the data are masked without releasing private information?

dzhuang:
>>What asserssions were supposed to be made?

> This is intended as a combined administrative/self-service function, with the main purpose being that the site admin would email course owners about this data being deleted, and after a week going ahead and doing so. Alternatively, the admin could ask course owners to do so themselves. The functionality should support both models.

Sorry, by `asserssion`, I mean in terms of tests.

inducer:
> Got it. But that is required even the data are masked without releasing private information?

Yep. https://oprs.research.illinois.edu/regulations-policies/university-policies

dzhuang:
> Yep. https://oprs.research.illinois.edu/regulations-policies/university-policies

That's cool...

dzhuang:
http://docs.celeryproject.org/en/latest/reference/celery.contrib.testing.worker.html#celery.contrib.testing.worker.start_worker.

For integration test,  `start_worker` can be used for testing celery tasks. However, it's only available for `celery >=4`. Or else we can only do unittest using `mock`.

dzhuang:
However, Celery 4 do not support Windows, that makes it impossible, as least for me, to continue developping  and testing locally.

dzhuang:
That may not be the end of the world, there is https://github.com/celery/celery/issues/4082. The last thing to determine if we need to drop djcelery is, how valuable is djcelery's admin utility for RELATE (and potential backward compatibility issue)?

inducer:
>  The last thing to determine if we need to drop djcelery is, how valuable is djcelery's admin utility for RELATE (and potential backward compatibility issue)?

I've never used it. I doubt anyone else has on my end. In addition, the replacement (`django-celery-results`) comes with a new admin thing that looked smaller and more useful. (FWIW, I *have* used that.)

I'm not sure there's much of a migration concern, as long as everyone makes sure their task queues are empty before upgrading. Maybe we should start doing releases and release notes to highlight stuff like this.

dzhuang:
There's also a task monitor which doesn't support by 4, and that might be a reason for staying at 3?

dzhuang:
As for #433 , testing celery is not that painful as I had expected.

inducer:
> might be a reason for staying at 3

Celery 3 has been "old and busted" with Django warnings for a while. I don't really want to stick with it.

The biggest downside of Celery 4 to me is that the Django DB can no longer serve as a task broker, thus requiring another outside dependency. But even that is not enough to make me want to stay with 3.

inducer:
@dzhuang Does this look good to you?

dzhuang:
> Does this look good to you?

I think it's ok.

inducer:
In it goes.

PR #415, opened 2018-01-15
Fix template overriding continued

dzhuang:
Add back {% block header_extra %}{% endblock %} to base.html

### DIFF
--- a/relate/templates/base.html
+++ b/relate/templates/base.html
@@ -35,6 +35,8 @@
         {{ form.media }}
       {% endblock %}
 
+      {% block header_extra %}{% endblock %}
+
       {% block le_ie9_fallback %}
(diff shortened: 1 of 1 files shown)


dzhuang:
Todo: Add test the ensure those assets are included in specific views.

inducer:
Is this ready to merge, despite the "Todo" comment? If not, please add [WIP] to title.

dzhuang:
It's ready, but I will add tests later.

inducer:
Sounds good. Thanks for the quick response (and the patch)!

dzhuang:
Off the topic. (I always try to respond in time. But you know I'm using Gmail which is banned here (that posed some lag) and there's timezone between us.)

PR #412, opened 2018-01-15
[WIP] Towards using webpack, discontinue django-bower

inducer:
cc @dzhuang 

This was long overdue. Bower is showing its age, plus django-bower is showing issues like https://github.com/inducer/relate/commit/c57e4a90831e1b5ddbde4f95750efd50333cef6e#commitcomment-26845580. I imagine it'll help quite a bit with our page load times, too.

PR #419, opened 2018-01-15
Make choice question data handling less string-y

inducer:
@dzhuang Could you give this a look, since it's an important code path?

### DIFF
--- a/course/page/choice.py
+++ b/course/page/choice.py
@@ -24,7 +24,7 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 THE SOFTWARE.
 """
 
-
+import six
 from six.moves import range
 import django.forms as forms
 from django.utils.safestring import mark_safe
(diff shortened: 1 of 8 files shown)


dzhuang:
I'll test that.

course/page/choice.py
```diff
@@ -65,20 +64,78 @@ def markup_to_html_plain(page_context, s):
...
 
-# {{{ choice question base
+# {{{ choice data model
 
-class ChoiceQuestionBase(PageBaseWithTitle, PageBaseWithValue):
+class ChoiceModes:
+    INCORRECT = "incorrect"
+    CORRECT = "correct"
+    DISREGARD = "disregard"
+    ALWAYS_CORRECT = "always_correct"
```
dzhuang: Can you explain the use case of "aways_correct"? Thanks.

inducer: Sure. It exists to ensure fair scoring of a multi-select question in which one option has turned out to be flawed. The net effect of `disregard` is to score the question as if that option didn't exist. But some students may have received points from the broken option, so `disregard` would take those points away. Cue lots of (somewhat justified) complaints from grumpy students. `always_correct` prevents that by grading any answer as a correct one, therefore never leading to a point decrease.

If this explanation was helpful, could you copy-paste it to the docs? Thanks! :)

dzhuang: Ok. Thanks.

dzhuang: I've submit a PR in https://github.com/inducer/relate-sample/pull/10/files, can you review that?

course/page/choice.py
```diff
@@ -65,20 +64,78 @@ def markup_to_html_plain(page_context, s):
...
+                            _("%(location)s: more than one choice mode set")
+                            % {'location': location})
+
+                item_mode[0] = mode
+
+            return s
+
+        s = mode_from_prefix(ChoiceModes.CORRECT, cls.CORRECT_TAG, s)
+        s = mode_from_prefix(ChoiceModes.DISREGARD, cls.DISREGARD_TAG, s)
+        s = mode_from_prefix(ChoiceModes.ALWAYS_CORRECT, cls.ALWAYS_CORRECT_TAG, s)
```
dzhuang: This won't handle `~CORRECT~~CORRECT~foobar`, i.e., it will view `~CORRECT~foobar` as a correct answer.

inducer: Should it?

dzhuang: (sorry for late reponse) `~CORRECT~~DISREGARD~foobar` will raise `ValidationError`, am I right?

dzhuang: And `~DISREGARD~~CORRECT~foobar` is also expected to raise a `ValidationError` but not.

inducer: Agreed,

inducer:
There are some merge conflicts in the tests. @dzhuang, could you take a look?

dzhuang:
@inducer To resolve this, you need to merge #442 into this PR. and merge #441 into master. Then it can be resolved.

inducer:
@dzhuang, you mentioned in #412 that it LGTY, so I'll go ahead and merge it.

dzhuang:
Thank you！

PR #421, opened 2018-01-15
Flow session contents API

inducer:
@davis68 This along with #419 should help with what you are trying to do.


inducer:
It also adds some minimal API docs, based on your, um, suggestion. :)

inducer:
Looks like this depends on #418 after all.

PR #422, opened 2018-01-15
[WIP] Allow upgrade to Django 2.0

inducer:
cc @dzhuang 

Also jump onto new version of Celery.

Things to resolve:

- [x] Django 2.0 drops Py2
- [x] Update install docs: Celery 4 requires a message broker, e.g. redis or rabbitmq


inducer:
@dzhuang I'll wait with this until you have a chance to test it/give the go-ahead.

inducer:
The unicode fixes in ce58e1b are probably worth having no matter what.

inducer:
Heh. Kind of forgot about the biggest part of this. Django 2.0 is Py3 only. Personally, I've been on Py3 in production for a while, and I'm ready to let Py2 go.

Either way, we would need some CI engineering.

How about you?

dzhuang:
Got it.

dzhuang:
Can you merge master into this? Thanks.

inducer:
Done.

inducer:
Done.

inducer:
@dzhuang How do you feel about this? I'd prefer to get this in this weekend.

For now, we're just *allowing* Django 2, so it doesn't rule out Py2 just yet. Plus my Celery is slightly broken in production, and I'd like to only re-debug it once. (rather than twice--now and again when we hop to Celery 4.)

inducer:
> Either way, we would need some CI engineering.

I no longer believe that's true. Py2 will just select Django 1.11 automatically.

dzhuang:
I'm ok with this, but the docs were not ready yet. 

Sorry for late response (until I finally found a way to access my gmail).

inducer:
@dzhuang I'm using this brach to also switch to pipenv (which I seem to rather like for predictable deployment). Could you give this a try and see what you think?

dzhuang:
You can merge it once you think it's ready (I won't have time to test this in a couple of hours). I'll complain if it failed for me. :)

inducer:
OK, thanks! I'll at least make sure I keep the CIs working.

inducer:
Wow, seriously, forget Pipenv. What a waste of time.

dzhuang:
I see there is a long que in appveyor ci (due to hang of previously failed build). I think you can manually cancel those build to see whether the latest build works.

inducer:
In it goes.

dzhuang:
Finally, thanks. Let's wait and see :)

Issue #417, opened 2018-01-15
Bug: Async bulk regrade (need confirm)


dzhuang:
When the grading strategy is `using_latest`, grades of participations with multiple sessions will using the session which started first, not the expected latest.

inducer:
Hum. This one is actually tricky to fix in a backwards-compatible manner. What I view as the correct behavior would be to set the grade time of the grade change created [here](https://github.com/inducer/relate/blob/444dcd2f9780cce236d5c019a25ba876f1278479/course/flow.py#L1123) to the completion time of the flow.

The problem with that is that it will place new grade changes created using this behavior *before* grade changes created earlier using the broken behavior. A clunky but effective way to deal with this is to delete all the old grades produced by this flow. But I don't see any solution that's substantially better.

What do you think?

dzhuang:
Will extending #264 (closed)  solving the both? 

The problem with #264 is it didn't consider `GradeChange` objects which do not have a session attribute, as you commented, by that is easy to modify, by adding a branch.

The logic is, the grade should be determined by submission time, not grade time, for flows using time-based strategy.  That should apply for both this issue and #263.

dzhuang:
What I said above is not correct, if consider a session which have both imported grade and form submitted grade.

inducer:
> Will extending #264 (closed) solve both?

I believe that's possible. It also neatly solves the backward compatibility concern. I'm in favor.

PR #423, opened 2018-01-16
Fix #263 and #417

### LINKED ISSUE #263: Bug: wrong overall grade when grading criterion changed between multiple submissions

dzhuang:
Not tested yet.


course/models.py
```diff
@@ -1587,6 +1587,23 @@ def get_state_desc(self):
         return dict(GRADE_STATE_CHANGE_CHOICES).get(
                 self.state)
 
+    # Compare GradeChange objects in terms of time, in order to correctly sort
+    # objects, fixing #236 and #417
+    def __eq__(self, other):
+        if not (self.flow_session and other.flow_session):
+            # at least one object has flow_session, then compare them by grade_time
```
dzhuang: This should be "at least one object doesn't have flow_session"

course/models.py
```diff
@@ -1587,6 +1587,23 @@ def get_state_desc(self):
...
+            return self.grade_time == other.grade_time
+        else:
+            # if both objects have a flow_session
+            return self.flow_session.completion_time == other.completion_time
+
+    def __gt__(self, other):
+        if not (self.flow_session and other.flow_session):
+            return self.grade_time > other.grade_time
+        else:
+            return self.flow_session.completion_time > other.completion_time
```
inducer: I think this might not yield a consistent ordering in the presence of grade changes with and without flow sessions.

Better:

```
self_time = self.grade_time
if self.flow_session is not None:
    self_time = self.flow_session.completion_time
other_time = ...
```

Same for `__eq__`. Maybe this time getting should be bagged up in its own function. There is also a potential performance concern, since `self.flow_session.completion_time` involves another query unless the flow sessions were brought in with `select_related` or some such. So we should comb through the usage sites of this and make sure that's the case.

dzhuang: My idea was the compare should be determined by which objects we are compare. (Selecting key according to which type of Gradechange objects we are comparing). 

1. If either objects have no flow_session, we use grade_time as key.

1. If both have flow_session, we use flow_session.completion_time as key.

I don't think  `self_time` can achieve that, because you don't know yet who is going to be compared.

dzhuang: There was a mistake:
```
return self.flow_session.completion_time > other.completion_time
```
should be:
```
return self.flow_session.completion_time > other.flow_session.completion_time
```

course/models.py
```diff
@@ -1587,6 +1587,23 @@ def get_state_desc(self):
...
 
+    # Compare GradeChange objects in terms of time, in order to correctly sort
+    # objects, fixing #236 and #417
+    def __eq__(self, other):
+        if not (self.flow_session and other.flow_session):
+            # at least one object has flow_session, then compare them by grade_time
+            return self.grade_time == other.grade_time
+        else:
+            # if both objects have a flow_session
+            return self.flow_session.completion_time == other.completion_time
```
dzhuang: return self.flow_session.completion_time == other.flow_session.completion_time

course/models.py
dzhuang: > There is also a potential performance concern, since self.flow_session.completion_time involves another query unless the flow sessions were brought in with select_related or some such. So we should comb through the usage sites of this and make sure that's the case.

That's a good point.

dzhuang: > I think this might not yield a consistent ordering in the presence of grade changes with and without flow sessions.

Can you give an example?

About inconsistency might happen if two GradeChange objects have the same time attribute (we selected, though it's rare) while with different percentage value. So I'll add pk if that happens.

dzhuang: (Then `__eq__` is meaningless.)

inducer:
Still WIP?

dzhuang:
Yes. I planed to add tests for this tomorrow.

course/models.py
inducer: > Can you give an example?

I'll try:
```
A: grade time 2015
B: grade time 2016, flow completed 2014
C: grade time 2017
D: grade time 2018, flow completed 2013
```

Then `A < B < C`

But also `D < B` and `C < D`, which is a direct contradiction.

dzhuang: That's interesting. I'll think about that tomorrow.

dzhuang: Since that inconsistency exists (as you had demostrated), defining `__gt__` is not a good work around, at least it might be wrongly used in the future, I think it's better to implement the sorting within the range of `consume` method.

dzhuang: Both previous commits will fail when reopen a graded (finished session). Because those GradeChange objects has a flow_session, but that session don't have a completion_time now.

dzhuang:
This PR should also work round #430 .

dzhuang:
If `use_latest`, a latest session which is sumbitted but not graded, the overall grade should be (None), not the last second session's grade.

course/models.py
inducer: > Since that inconsistency exists (as you had demostrated), defining `__gt__` is not a good workaround, at least it might be wrongly used in the future

It's perfectly possible to define `__gt__` in a manner that doesn't run into inconsistencies. See my first comment with the `self_time`.

I guess I'll spell out the full set of methods that I think makes sense:

```python
def sort_time(self):
    return self.flow_session.completion_time if self.flow_session is not None else self.grade_time

def __lt__(self, other):
    if not isinstance(other, GradeChange):
        return NotImplemented

    return self.sort_time() < other.sort_time()

def __le__(self):
    if not isinstance(other, GradeChange):
        return NotImplemented

    return self.sort_time() <= other.sort_time()

def __gt__(self):
    if not isinstance(other, GradeChange):
        return NotImplemented

    return self.sort_time() > other.sort_time()

def __ge__(self):
    if not isinstance(other, GradeChange):
        return NotImplemented

    return self.sort_time() >= other.sort_time()
```

dzhuang: That sounds good. I have finally done with my grading work of the ended semester. And I'll look into this today.

dzhuang: For this part
```python
def sort_time(self):
    return self.flow_session.completion_time if self.flow_session is not None else self.grade_time
```

For some GradeChange objects, it does have a `flow_session`, but when that session is reopened, then `completion_time` is None, which is not comparable and the sorting will fail.

dzhuang: And, we should not implement a method which can give an equal compare. Because that's another source of inconsistency. so `__ge__` and `__le__` should be removed. When `sort_time()` return two equal result, we should use only the one which comes later, IOW, which has greater `pk`.

dzhuang:
I think we'll have to first try to enable tests of celery utility on CIs before we can continue this PR.

course/models.py
inducer: > For some GradeChange objects, it does have a flow_session, but when that session is reopened, then completion_time is None, which is not comparable and the sorting will fail.

Fair. So, catch that, and treat that case as if no flow session existed?

inducer: >  so __ge__ and __le__ should be removed.

No objection. What methods does Python's sorting use?

dzhuang: > No objection. What methods does Python's sorting use?

Simply `gt` might be enough.

inducer: You know, I think you're right. If we can't supply a full suite of comparison operators, then maybe this is better handled by a `key=` function handed to `sort`.

inducer:
#263 and #417 (for links)

course/grades.py
```diff
@@ -792,6 +796,26 @@ def finalize():
...
+        .select_related("participation__user")
+        .select_related("creator")
+        .select_related("flow_session")
+        .select_related("opportunity"))
+
+    state_machine = GradeStateMachine()
+    state_machine.consume(grade_changes, set_is_superseded=True)
+
+    return grade_changes, state_machine
+
```
dzhuang: I refactor this out for easier (and faster) tests without client.

course/migrations/0112_add_previous_completion_time_to_flow_session.py
```diff
@@ -0,0 +1,34 @@
...
+        migrations.AddField(
+            model_name='flowsession',
+            name='previous_completion_time',
+            field=models.DateTimeField(blank=True,
+                                       help_text='The last completion time for an in-progress flow session, which had been submitted and then reopened.',
+                                       null=True,
+                                       verbose_name='Previous completion time'),
+        ),
+        migrations.RunPython(set_default_previous_completion_time)
+    ]
```
dzhuang: This is a shorthand way for judging whether a session is reopened, and make it easier for sorting GradeChanges with a session.

course/models.py
```diff
@@ -1642,8 +1652,8 @@ def _consume_grade_change(self, gchange, set_is_superseded):
 
         # check that times are increasing
         if self._last_grade_change_time is not None:
-            assert gchange.grade_time > self._last_grade_change_time
-            self._last_grade_change_time = gchange.grade_time
+            assert gchange.grade_time >= self._last_grade_change_time
+        self._last_grade_change_time = gchange.grade_time
```
dzhuang: Wrong indentation of the original code?

course/models.py
```diff
@@ -1704,11 +1716,55 @@ def consume(self, iterable, set_is_superseded=False):
...
+                    flag = -2
+                elif gchange.grade_time == time_pre_completed_sessions:
+                    flag = -1
+                elif gchange.grade_time == time_post_completed_sessions:
+                    flag = 1
+                elif gchange.grade_time > time_post_completed_sessions:
+                    flag = 2
+                else:
+                    flag = 0
+                gchange_flag_tuple.append((gchange, flag))
```
dzhuang: I added an axes to compare GC objects. We don't need full sort, just return a list with a correct first and last GC is ok.

course/models.py
```diff
@@ -1704,11 +1716,55 @@ def consume(self, iterable, set_is_superseded=False):
...
+
+            gchange_sorted = sorted(gchange_flag_tuple,
+                                     key=lambda x: (x[1], x[0].pk))
+
```
dzhuang: With pk as second sort key.

dzhuang:
Updated. I think the sorting part (in models modual) is not Pythonic (IOW, ugly :-( ).

dzhuang:
I added a grade change state named "session_reopened", to deal with #430.

dzhuang:
This PR is ready for review, thanks @inducer

dzhuang:
I'll add more tests.

course/models.py
dzhuang: As for you example:
```
A: grade time 2015
B: grade time 2016, flow completed 2014
C: grade time 2017
D: grade time 2018, flow completed 2013
```
If `use_earliest`, then A will be used.
If `use_latest`, then C will be used.

This is tested in the latest commit. The logic is simpler, just keep the GradeChange object with the latest completed session when sorting. This mean that we need to use the grading strategy when consuming.

dzhuang:
This PR is ready for review, thanks @inducer

course/models.py
```diff
@@ -1704,16 +1721,60 @@ def consume(self, iterable, set_is_superseded=False):
...
+        import copy
+        _valid_percentages_copy = copy.copy(self.valid_percentages)
 
         self.valid_percentages.extend(
                 cast(GradeChange, gchange.percentage())
                 for gchange in valid_grade_changes)
 
+        # {{{ Calculate the earliest percentage and latest percentage, instead of
+        # get them from self.valid_percentages, so as to avoid inconsistent results.
+        # See https://github.com/inducer/relate/pull/423#discussion_r162121467
```
inducer: That link does not work.

course/constants.py
```diff
@@ -504,6 +504,7 @@ class grade_state_change_types:  # noqa
     report_sent = "report_sent"
     do_over = "do_over"
     exempt = "exempt"
+    session_reopened = "session_reopened"
```
inducer: How is `session_reopened` different from a do-over?

course/migrations/0112_add_previous_completion_time_to_flow_session.py
inducer: > make it easier for sorting session-related GradeChanges.

To my mind, you should be able to evaluate a grade without having to bring in any associated `FlowSession`s.

inducer: I think a better path for this would be to have an `effective_time` and a `creation_time` for `GradeChange`s. The `effective_time` would then match the last completion time of an event.

course/models.py
dzhuang: Oh, that link is targeting our previous discussion on the potential inconsisitency might araise when sorting `GradeChange` objects with/without a session.

course/constants.py
dzhuang: If `do_over` is the latest grade change, it will clear previous grades, with stringified state is `NONE`. If  `session_reopened` as just like `do_over`, but with stringified state `OTHER_STATE` when it is the last grade change.

course/migrations/0112_add_previous_completion_time_to_flow_session.py
dzhuang: As discussed in #430, we need student/grader to know the latest session is reopened, there will be no grade for the opportunity before the reopened session is resubmitted and the resubmitted is graded.

To achieve that, we need to differentiate in-progress sessions between ones which has previously been submitted and then reopened, and ones which hasn't. That can be done by introduction `previous_completion_time` field, because for in-progress sessions which had never been submitted, that field is None. 

This also make it possible to find out which session is the actually the latest session (for grading strategy "use_latest"). Because (IMO and you might also agree), the latest is the session which is submitted latest. Without this, when a latest session is reopened, it is not the latest session any longer. That's a wrong behavior.

course/models.py
dzhuang: (The discussion is collapsed, and that link can't uncollpase it).

course/migrations/0112_add_previous_completion_time_to_flow_session.py
dzhuang: I used it here https://github.com/inducer/relate/pull/423/files#diff-5a7d8b42274c6175f980693ada70368bR1749

dzhuang: What do you think about this? I think this issue is somewhat urgent for me.

dzhuang: Can you reply on this discussion? @inducer 
This fix is quite urgent for me. If that bug continued without fixing, some students will definitely complain about their final points. And I won't have time to do more coding.

> I think a better path for this would be to have an effective_time and a creation_time for GradeChanges. The effective_time would then match the last completion time of an event.

Since you seemed to agree on #417 , without adding a `previous_completion_time` field into `Flowsession`, how can we know the session is a reopened one (instead of a just-started on ), and the GradeChange is from a reopened session?

dzhuang: I just realized that finishing a session will create a gradechange object, whether autograded or not. I now think your proposal is more reasonable than mine.

PR #429, opened 2018-01-17
Fix #428

### LINKED ISSUE #428: Add warning for browsers <= IE 9
IE8 is not working well with HTML5, for example, bootstrap dropdown, collapsible and accordion elements won't respond to user interactions, and videojs failed to load for each page. Thus, simply fallback with https://github.com/inducer/relate/blob/master/relate/templates/base.html#L41-L45 doesn't make sense. We'd better add warning about that (and suggest using modern browser) in template, and it's easy.

dzhuang:


relate/templates/base.html
```diff
@@ -59,6 +51,13 @@
   </head>
 
   <body>
+    {% block lte_ie9_warning %}
+      <!--[if lte IE 9]>
+        <div class="browser alert alert-danger warning"><i class="fa fa-warning"></i>
+          {% trans "This website may not be compatible with your outdated Internet Explorer version. If you want use Internet Explorer, please install one with version higher than 10." %}
```
inducer: "higher than 10" -> "10 or higher"?

dzhuang: You're right. Can you correct that? thanks.

inducer:
Done

relate/templates/base.html
inducer: Done

dzhuang: Thank you!

inducer:
Thx!

PR #431, opened 2018-01-19
Allow open local_settings.py with non-ASCII characters on Windows.

dzhuang:
In this way, I can write Chinese character in `RELATE_SITE_ANNOUNCEMENT` in local_settings.py. And, enabled auto-restarting (developing) server when local settings changes.


relate/settings.py
```diff
@@ -306,4 +306,14 @@
 
 # }}}
 
+# {{{ Enable auto reload when local_settings changes when debugging
+
+if globals().get("DEBUG", False):
+    try:
+        import local_settings  # type: ignore
```
inducer: Why not replace the `open`/`exec` logic above with an import?

dzhuang: I don't know that's also OK.

inducer: > I don't know that's also OK.

Why wouldn't it be? It requires a little rewiring, but it's probably better anyhow.

inducer: The one issue perhaps is to make sure that it can be found by the `import`.

dzhuang: > The one issue perhaps is to make sure that it can be found by the `import`.

Yes. I don't know how to make it working in that way. We need to import the local settings by name, since it doesn't have a consistent name for runserver or runtests.  `importlib.import_module` can import a module by name, but with the prerequisit that the dirctory has an `__init__.py`, which is missing in our case.

dzhuang:
Sorry, because conflicts, I squashed this PR. The latest commit used `exec` to do the import.

dzhuang:
Since we can't do
```python
import local_settings.example
```
I had to rename "local_settings.example.py" to "local_settings_example.py".

dzhuang:
This PR is ready for review.

relate/settings.py
```diff
@@ -14,9 +14,11 @@
 from django.conf.global_settings import STATICFILES_FINDERS, gettext_noop
 
 # Build paths inside the project like this: os.path.join(BASE_DIR, ...)
+import sys
 import os
 from os.path import join
 BASE_DIR = os.path.dirname(os.path.dirname(__file__))
+sys.path.append(BASE_DIR)
```
inducer: Can we do without this?

relate/settings.py
```diff
@@ -27,17 +29,22 @@
...
+local_settings_module_name = (
+    os.path.splitext(os.path.split(_local_settings_file)[-1])[0])
+exec("import %s as local_settings" % local_settings_module_name)
+
+# Suppress syntax warning in following scripts
 try:
-    with open(_local_settings_file) as inf:
-        local_settings_contents = inf.read()
-except IOError:
+    local_settings = local_settings  # type: ignore  # noqa
```
inducer:
```
local_settings = local_settings_module.__dict__
```
? And then back out all the `getattr` changes below?

relate/settings.py
```diff
@@ -27,17 +29,22 @@
...
-        }
+if not os.path.isfile(_local_settings_file):
+    raise RuntimeError(
+        "Management command '%(cmd_name)s' failed to run "
+        "because '%(local_settings_file)s' is missing."
+        % {"cmd_name": sys.argv[1],
+           "local_settings_file": _local_settings_file})
+
+local_settings_module_name = (
+    os.path.splitext(os.path.split(_local_settings_file)[-1])[0])
```
inducer: Assert that the file extension is `.py`?

relate/settings.py
dzhuang: Can you spell it out? :-)

relate/settings.py
dzhuang: That's a good idea.

relate/settings.py
dzhuang: That's smart.

relate/settings.py
dzhuang: Opooos, I didn't know that's possible....

dzhuang:
Updated, thanks for the smart suggestions.

relate/settings.py
```diff
@@ -27,17 +28,19 @@
...
+    raise RuntimeError(
+        "Management command '%(cmd_name)s' failed to run "
+        "because '%(local_settings_file)s' is missing."
+        % {"cmd_name": sys.argv[1],
+           "local_settings_file": _local_settings_file})
+
+local_settings_module_name, ext = (
+    os.path.splitext(os.path.split(_local_settings_file)[-1]))
+assert ext == ".py"
+exec("import %s as local_settings_modual" % local_settings_module_name)
```
inducer: `modual`: Typo?

dzhuang: Yeap. I'll correct that.

dzhuang: updated.

dzhuang:
I think this PR is ready now.

dzhuang:
I think this PR is ready now. @inducer

PR #433, opened 2018-01-19
Add tasks tests.

dzhuang:


course/flow.py
```diff
@@ -1017,10 +1017,11 @@ def expire_flow_session(
...
 
-    if (past_due_only
-            and grading_rule.due is not None
-            and now_datetime < grading_rule.due):
-        return False
+    if past_due_only:
+        if grading_rule.due is None:
+            return False
+        elif now_datetime < grading_rule.due:
+            return False
```
dzhuang: Do you think if due is None, then never due is a better behavior?

inducer: I think that's what that implements, no?

dzhuang: No,  when doing bulk expire in progress session.

inducer: Oh, whoops. Half-asleep apparently. I looked at the code with the green background. :)

Yeah, that seems fine semantically.

dzhuang: : -). NVM.

course/flow.py
```diff
@@ -1243,10 +1244,10 @@ def finish_flow_session_standalone(
...
 
-    if grading_rule.due is not None:
-        if (
-                past_due_only
-                and now_datetime_filled < grading_rule.due):
+    if past_due_only:
+        if grading_rule.due is None:
+            return False
+        elif now_datetime_filled < grading_rule.due:
             return False
```
dzhuang: Same as above.

dzhuang:
I am using `@override_settings(CELERY_TASK_ALWAYS_EAGER=True)` to fake the async operation, that made the assumption there are no serialization issues.

inducer:
Awesome. Thank you for working on this.

PR #436, opened 2018-01-23
Fix Django 2 compatibility.

dzhuang:
@inducer I think it's better to put the PR here. It passed tests on Django 1.1x and Celery 3.1. 

I will be working on the tests of Celery 4. When it's ready, you can merge it into #422 .


inducer:
Thanks for working on cleaning all this up!

> Then we can have 2 versions (the difference is on the depedencies and configurations of Celery).

Two versions of what? Of all of Relate? That seems way hard.

dzhuang:
> Two versions of what? Of all of Relate? That seems way hard.

Ok. So we won't do that.

dzhuang:
Done. 
BTW, tests are added to ensure serialization of celery tasks.

dzhuang:
Tests for Python2 is preserved, with highest version of Django < 2, and passed tests for celery 4.

inducer:
Thanks for your work on this!

PR #438, opened 2018-01-23
Remove use of dnspython3, which is no longer necessary

inducer:


inducer:
Ref: https://pypi.org/project/dnspython3/

PR #441, opened 2018-01-31
Refactor sandbox and some tests.

dzhuang:
This is used to resolve conflicts, should be merged before merging #419.


dzhuang:
@inducer This also need to be merged. Actually, the change in sandbox.py is copied from what I've done in #419. I found no way to resolve that without write access.

inducer:
Thx!

PR #443, opened 2018-01-31
Avoid failure when student passes non-callable as function

bowmanat:
When students pass a non-function as `f` and we use `feedback.call_user(f)`, the grading code fails when trying to get `f.__name__`.  This patch resolves the issue by checking whether `f` is a `callable`.

One thing to note:  `callable()` was removed in Python 3.0 and 3.1, but brought back in later versions.  If you do not want to use `callable()`, you could accomplish the same thing with 

`import collections`
`isinstance(f, collections.Callable)`

I have *not* been able to test this on a running RELATE instance, so please look it over closely.

### DIFF
--- a/course/page/code_feedback.py
+++ b/course/page/code_feedback.py
@@ -171,6 +171,14 @@ class Feedback:
             return f(*args, **kwargs)
         except Exception as e:
             if callable(f):
+                try:
+                    callable_name = f.__name__
+                except Exception as e_name:
+                    callable_name = (
+                                "<unable to retrieve name; encountered %s: %s>"
+                                % (
+                                    type(e_name).__name__,
+                                    str(e_name)))
                 from traceback import format_exc
                 self.add_feedback(
                         "<p>"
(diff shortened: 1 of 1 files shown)


inducer:
Thanks for the patch! I agree with your assessment that the current code could be made more robust.

Actually, not everything that's callable has a `.__name__`, so I think your patch might be improved by simply guarding the `.__name__` query, like so:

```python
try:
    callable_name = f.__name__
except Exception as e:
    callable_name = "<unable to retrieve name; encountered %s: %s>" % (type(e).__name__, str(e))
```

bowmanat:
I agree that your patch addresses the problem more directly and is probably a good way to go.  However, I think checking for callable gets more at the cause and less at the symptom, and it allows for a more readable error message for the students.

How would you feel about having both patches -- first, tell the student if they failed to submit a function, and second, add the try/except on `f.__name__` to make the code more robust?

inducer:
Both sounds good. Thanks for working on this!

inducer:
Tested, looks good. Thanks for your contribution!

PR #446, opened 2018-02-01
Improve list_filter in admin to show only objects from related course.

dzhuang:


inducer:
Although I think in the long run our only option is to stop using the admin (because it can't be secured), this is a nice improvement if it works as advertised.

dzhuang:
> Although I think in the long run our only option is to stop using the admin.

I agree, since currently many model CRUD are not available, I used that as an alternative for some instructors.

dzhuang:
This also fixed #447 .

inducer:
This looks good. Plus it's easily backed out if it happens to break something.

dzhuang:
Thanks for the feedback. Then I'll merge it.

PR #448, opened 2018-02-05
Fix bug when feedback points are close enough to 1

adamjstewart:
We have a lot of multi-part homework questions. We initialize a score of 0, then add points for each part the student gets correct. For example:
```python
score = 0

if x is good:
    score += 0.3
if y is good:
    score += 0.3
if plot is good:
    score += 0.2
if plot has a title:
    score += 0.05
if plot has an xlabel:
    score += 0.05
if plot has a ylabel:
    score += 0.05
if plot has the correct axes:
    score += 0.05

feedback.set_points(score)
```
This test code looks innocuous enough, until you realize that `score` doesn't add up to 1, it adds up to 1.0000000000000002. This results in RELATE crashing with the following error message:
```
  File "/opt/runpy/runpy", line 89, in do_POST
    run_code(response, run_req)
  File "/opt/runpy/code_runpy_backend.py", line 303, in run_code
    % feedback.points)
ValueError: grade point value is invalid: 1.0000000000000002
```
To get around this, we've been adding the following code snippet to all of our test codes:
```python
import math

if math.isclose(score, 1):
    feedback.set_points(1)
else:
    feedback.set_points(score)
```
I figured it was time I got around to reporting and fixing this bug.

**Note:** I haven't actually tested this change. I have no idea how to test this.

This is one of many ways to solve the problem. Personally, I think it would be easier to accept any value greater than 1 as 1, so if the point total can go over 100%, let it. I wasn't sure if it would be better to catch the bug in this file, or in code_feedback.py. Personally, I would use a setter for the points attribute that checks these kind of things.

### DIFF
--- a/course/page/code_runpy_backend.py
+++ b/course/page/code_runpy_backend.py
@@ -24,6 +24,7 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 THE SOFTWARE.
 """
 
+import math
 import sys
 import traceback
 
@@ -299,6 +300,9 @@ def run_code(result, run_req):
         except Exception:
             package_exception(result, "test_error")
             return
+
+    if feedback.points is not None and math.isclose(feedback.points, 1):
+        feedback.points = 1
 
     if not (feedback.points is None or 0 <= feedback.points <= 1):
         raise ValueError("grade point value is invalid: %s"


inducer:
Thanks for your contribution! It turns out that the [actual criterion](https://github.com/inducer/relate/blob/74ee62b056538d188ae1cdbc7a4552ee9eff8efd/course/page/base.py#L193) is much less strict--it allows "correctness" values of up to 10. So I am guessing we could use that same value to test against. Ideally, we'd pass it into the docker container via JSON (as, say, `max_correctness`) and compare with that. (Unfortunately, we can't just import `course.constants` in the container.)

As for testing, clone relate and say `pip install -e .` and then run
```
relate test-code some/problem.yml
```

dzhuang:
@adamjstewart I also meet the issue you mentioned, and I used simlar workaround like yours. I think your patch is a valid one, and will be better if following the route suggested by @inducer .

For testing the changes, currently the file you patched is out of the scope of coverage (because that file is used in a subprocess when testing, and I did some try to combine it into the report following [this guild](http://coverage.readthedocs.io/en/coverage-4.2/subprocess.html) but failed. So we had to do docer related tests via integration test. For you patch, You can test it in sandbox, like what we have recently done in (choice question)[https://github.com/inducer/relate/blob/master/tests/test_pages/test_choice.py].

adamjstewart:
I'm starting to wonder if I've bit off more than I can chew. I'm not sure why you're testing the range of points in multiple places instead of a single place. Wouldn't it make sense to add a setter for points so it is impossible to assign an invalid value?

P.S. I realized that we should also be checking for values close to -0. For example, someone might reasonably assign an initial score of 1.0 and subtract floating point values for each incorrect section, resulting in -0.0000000000000002.

inducer:
> P.S. I realized that we should also be checking for values close to -0. For example, someone might reasonably assign an initial score of 1.0 and subtract floating point values for each incorrect section, resulting in -0.0000000000000002.

That's a fair point.

inducer:
> I'm starting to wonder if I've bit off more than I can chew. I'm not sure why you're testing the range of points in multiple places instead of a single place.

Relate's autograding code run inside of a Docker container. The Python process inside of that container communicates with the main Relate app server over HTTP. The container runner thingie and the app server don't currently share code. They communicate through a HTTP+JSON protocol, defined in `course/page/code.py`. So we can't really ship `AnswerFeedback` objects back and forth. Currently, the runner thingie tries and performs some validation of the point count, but I'm coming to the realization that it just really shouldn't, which avoids all the complication I allude to above. Just let the app server validate the point count and complain if there's a problem. Ta-dah, no more duplicated code.

> Wouldn't it make sense to add a setter for points so it is impossible to assign an invalid value?

I'm not much into mutable state. In general, most objects in Relate don't change after they're created, so validating in the constructor is enough.

dzhuang:
@adamjstewart I am writing a PR to patch and test this PR. However, we are using relate-sample repo for the tests, and a new page type is merged there, so you PR is goint to fail if not merging the current master. Can you merge master into this PR?

dzhuang:
@inducer I've tested and patched the PR. But due to the reason
>However, we are using relate-sample repo for the tests, and a new page type is merged there, so you PR is goint to fail if not merging the current master. 

I think it's better to merge this PR and then submit my PR. I'm merging this.

PR #452, opened 2018-02-07
Added max_auto_feedback_points for code question along with tests.

dzhuang:
Patching #448 

Introduced new attribute `max_auto_feedback_points` for code questions.

### DIFF
--- a/course/page/code.py
+++ b/course/page/code.py
@@ -428,7 +428,7 @@ class PythonCodeQuestion(PageBaseWithTitle, PageBaseWithValue):
 
     .. attribute:: max_auto_feedback_points
 
-        Optional, A :class:`float` a non-negative value, default to 1,
+        Optional, A non-negative :class:`float` value, default to 1,
         and it can be increased to 10 (`course.constants.MAX_EXTRA_CREDIT_FACTOR`).
         It indicates the maximum allowed correctness for auto feedback.
 
--- a/course/page/code.py
+++ b/course/page/code.py
@@ -426,6 +426,12 @@ class PythonCodeQuestion(PageBaseWithTitle, PageBaseWithValue):
         based on its :attr:`access_rules` (not the ones of the flow), a warning
         is shown. Setting this attribute to True will silence the warning.
 
+    .. attribute:: max_auto_feedback_points
+
+        Optional, A :class:`float` a non-negative value, default to 1,
+        and it can be increased to 10 (`course.constants.MAX_EXTRA_CREDIT_FACTOR`).
+        It indicates the maximum allowed correctness for auto feedback.
+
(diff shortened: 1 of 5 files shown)


dzhuang:
@adamjstewart

dzhuang:
The tests on Travis-CI failed, but passed on Appveyor. I think the reason is the faked subprocess in the tests is not killed after one Testcase which opened that process.

dzhuang:
Confirmed. [this test](https://travis-ci.org/dzhuang/relate/jobs/338392163#L1290) is using the subprocess, and the TestCase followed that failed.

dzhuang:
@adamjstewart I also added test you mentioned in #448 that we should also make sure 0 is tested.
@inducer I am not confident about the naming and docs, can you review that? Thanks.

course/page/code.py
```diff
@@ -426,6 +426,12 @@ class PythonCodeQuestion(PageBaseWithTitle, PageBaseWithValue):
         based on its :attr:`access_rules` (not the ones of the flow), a warning
         is shown. Setting this attribute to True will silence the warning.
 
+    .. attribute:: max_auto_feedback_points
+
+        Optional, A non-negative :class:`float` value, default to 1,
+        and it can be increased to 10 (`course.constants.MAX_EXTRA_CREDIT_FACTOR`).
+        It indicates the maximum allowed correctness for auto feedback.
+
```
inducer: Do we really need that? What do you think of the simpler solution I proposed in #448 to not validate the point count in the runner and instead just do this in the app server?

dzhuang: I see this

> Ideally, we'd pass it into the docker container via JSON (as, say, `max_correctness`) and compare with that. 

So I guess that's your prefered solution. Thus I've made it more complex.

inducer: Yeah. I wrote that, and then last night I realize that I was being an idiot. No need to make it so complicated, IMO. Sorry!

dzhuang: Nvm, I started out without reading through your later comments. So the work around is remove that validation part in the docker side, and do it in the app server side?

dzhuang: You are right, It's better to do that in the app server, I'm closing this and open another PR.

dzhuang:
@inducer @adamjstewart When feedback.points falls out of the range of [0, MAX_EXTRA_CREDIT_FACTOR], besides `correctness = None` what is the expected behavior you prefer? 

1. Shall we log an error in the feedback
1. Shall we send an email to the instructor

inducer:
> Shall we log an error in the feedback?

Yeah. I'd suggest treating it the same as "grading code failed".

dzhuang:
Ok, that means we also need to send an email to the instructor. Thanks.

PR #453, opened 2018-02-07
Reduce test coverage value fluctuation.

dzhuang:
If test starts with TestCases which don't fetch the course content, the coverage be reduced, because we used cache to store the test course content, there will be no validation process for those tests.


dzhuang:
This is not working as expected, will be reverted.

PR #455, opened 2018-02-08
Fix bug when feedback points are close to Ubound and Lbound.

dzhuang:
Patching #448

### DIFF
--- a/course/page/base.py
+++ b/course/page/base.py
@@ -145,18 +145,18 @@ class InvalidFeedbackPointsError(ValueError):
     pass
 
 
-def get_close(value):
-    # type: (float) -> Union[float, int]
+def round_point_count_to_quarters(value, atol=1e-5):
+    # type: (float, float) -> Union[float, int]
     """
     If 'value' is close to an int, a half or quarter, return the close value,
     otherwise return the original value.
     """
 
-    if abs(value - int(value)) < ATOL:
+    if abs(value - int(value)) < atol:
         return int(value)
 
     import math
-    _atol = ATOL * 4
+    _atol = atol * 4
     v = value * 4
     if abs(v - math.floor(v)) < _atol:
         v = math.floor(v)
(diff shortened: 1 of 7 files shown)


dzhuang:
cc @inducer @adamjstewart

course/page/base.py
```diff
@@ -139,15 +139,57 @@ def markup_to_html(
...
 
+
+class InvalidFeedbackPointsError(ValueError):
+    pass
+
+
+CORRECTNESS_ATOL = 1e-5
+
+
+def get_close_value(correctness):
```
inducer: I'd move all this logic into one function, `validate_point_count`

course/page/base.py
```diff
@@ -139,15 +139,57 @@ def markup_to_html(
...
+    # type: (Optional[float]) -> (Optional[Union[float, int]])
+
+    if correctness is None:
+        return None
+    elif abs(correctness - 0) < CORRECTNESS_ATOL:
```
inducer: Systematically, we could maybe do this for every integer in the range up to `MAX_CREDIT_FACTOR`.

inducer: Maybe for halves and quarters, too--since they are exactly representable in FP.

course/page/base.py
```diff
@@ -139,15 +139,57 @@ def markup_to_html(
...
+            error_msg_pattern
+            % {"max_extra_credit_factor": MAX_EXTRA_CREDIT_FACTOR,
+               "invalid_value": correctness})
+    elif correctness == 0:
         return six.text_type(_("Your answer is not correct."))
-    elif abs(correctness - 1) < 1e-5:
+    elif correctness == 1:
         return six.text_type(_("Your answer is correct."))
     elif correctness > 1:
+        if correctness > MAX_EXTRA_CREDIT_FACTOR:
```
inducer: `+ ATOL` for consistency?

dzhuang: There's no need, because it had been rounded in `get_close_value`.

course/page/base.py
dzhuang: That's fair.

course/page/base.py
dzhuang: Good point!

dzhuang:
Updated.

dzhuang:
@inducer Can you confirm whether this is ok for merge. I think I need to use the newly created `get_close` function in work round grade change stringify. Thanks.

P.S., docker image update is needed, but I'm ok for not doing that in a while.

course/page/base.py
```diff
@@ -139,13 +140,63 @@ def markup_to_html(
 
 # {{{ answer feedback type
 
+
+class InvalidFeedbackPointsError(ValueError):
+    pass
+
+
+def get_close(value):
```
inducer: I'd prefer a more explicit name. `round_point_count_to_quarters`? Make `atol` a default argument rather than a global constant?

dzhuang:
Updated.

inducer:
LGTM

dzhuang:
Thank you for the review. Then I'm merging this, and you need to rebuild the image some day.

PR #456, opened 2018-02-09
[WIP] Optimize gradebook csv exportation

dzhuang:
- [ ] Add  tests.

Features: 
1. Allow define user prefered csv encoding. (For non latin language, can be open by excel without error).
1. Allow replace None in gradebook to 0.
1. Allow customized output fields in csv,
1. Allow enforce grades floor and ceiling for grades.
1. Enable round for grades.
1. Allow remove instructors and tas from the output.


dzhuang:
@inducer How do you feel about this PR? I am not confidence in the naming. Thanks.

course/grades.py
```diff
@@ -295,43 +308,195 @@ def grade_key(entry):
...
+            initial=100,
+            label=_("Maximum allowed points"),
+        )
+
+        self.fields["minimum_points"] = forms.FloatField(
+            required=True,
+            initial=0,
+            min_value=0,
+            max_value=100 * MAX_EXTRA_CREDIT_FACTOR,
+            label=_("Minimum points"),
```
inducer: Could use a help text. Don't know what it does.

course/grades.py
```diff
@@ -295,43 +308,195 @@ def grade_key(entry):
...
+        )
+
+        self.fields["round_digits"] = forms.IntegerField(
+            required=True,
+            min_value=0,
+            initial=0,
+            max_value=2,
+            label=_("Round Digits"),
```
inducer: Help text.

inducer: Or maybe change label to "Round point counts to N digits" and "Round percentages to N digits". Also, why max out at 2?

course/grades.py
```diff
@@ -295,43 +308,195 @@ def grade_key(entry):
...
+            required=False,
+            initial=True,
+            label=_("Exclude Grades of Teaching assistants"),
+        )
+
+        self.fields["zero_for_state_none"] = forms.BooleanField(
+            required=False,
+            initial=False,
+            label=_("Use 0 point for grade state `NONE`, which possibly "
+                    "means the opptunity had not been graded."),
```
inducer: Why "possibly"?

course/grades.py
```diff
@@ -295,43 +308,195 @@ def grade_key(entry):
...
+        self.fields["exclude_instructors"] = forms.BooleanField(
+            required=False,
+            initial=True,
+            label=_("Exclude Grades of Instructors"),
+        )
+
+        self.fields["exclude_tas"] = forms.BooleanField(
+            required=False,
+            initial=True,
+            label=_("Exclude Grades of Teaching assistants"),
```
inducer: "Teaching assistant" and "Instructor" are not concepts in Relate. That means you may not use them in any UI. (They used to be, but they no longer are.) Right now, there are just arbitrarily-defined roles. The only thing available is a "permission" that says whether a role's/participant's grades are to be included in the grades.

course/utils.py
```diff
@@ -1320,4 +1324,215 @@ def _render_notebook_from_source(
 
 # }}}
 
+
+class RelateCsvHandler(object):
+    # Tests ref tests.template_backends.test_django.DjangoTemplatesTests
```
inducer: I have no idea what this does and why it belongs here.

inducer: Why is this not in `checks.py`?

inducer: It could stand renaming to `RelateCSVSettingsValidator` or some such maybe?

course/grades.py
dzhuang: Ok.

course/utils.py
dzhuang: > I have no idea what this does and why it belongs here.
It's something that make the a csv configuration (currently from csv export) from settings.RelateCSVSetting, by check and add some default value if not configured, and packaged some methods like generate the user attribute choice fields (in the ExportCsvForm) for csv export.

dzhuang: > Why is this not in `checks.py`?

As mentioned above, it's functionalities are not limited to check only, do you prefer to put that in `checks.py`? Won't it look weird to import a function from `checks.py`?.

dzhuang: >It could stand renaming to RelateCSVSettingsValidator or some such maybe?

What about `RelateCSVSettingsInitializer`

course/grades.py
dzhuang: I previously set that to 4, and I found (in tests) the value are all rounded to 2 digits, had thought that's preset behavior. Since you commented on this,  I'll check that.

course/grades.py
dzhuang: That's why I opened #457 . I think it's quite confusing. both not-graded-yet and graded can result in `None`. Also, those participation who didn't open a session for the grade opportunity also get `None`.

dzhuang: I just opened an issue #461 about `NONE` .

course/utils.py
dzhuang: docstring added.

course/grades.py
dzhuang: All done.

dzhuang:
Updated.

course/grades.py
```diff
@@ -56,13 +57,15 @@
 from course.flow import adjust_flow_session_page_data
 from course.views import get_now_or_fake_time
 from course.constants import (
-        participation_permission as pperm,
+        participation_permission as pperm, MAX_EXTRA_CREDIT_FACTOR
         )
 
+csv_settings = RelateCSVSettingsInitializer()
```
inducer: Can we do without this global variable somehow?

course/grades.py
```diff
@@ -295,43 +309,207 @@ def grade_key(entry):
...
+                  "and \"Minimum percentage value\" below."))
+        )
+
+        self.fields["zero_for_graded_none"] = forms.BooleanField(
+            required=False,
+            initial=True,
+            label=_("0 percentage for graded 'NONE'"),
+            help_text=string_concat(
+                _("Use 0 percentage for graded opportunities which have "
+                  "`NONE` points."),
```
inducer: "`NONE` points" reads awkwardly. Maybe "that do not currently have a defined point count."

course/grades.py
```diff
@@ -295,43 +309,207 @@ def grade_key(entry):
...
+        )
+
+        self.fields["exclude_course_staff"] = forms.BooleanField(
+            required=False,
+            initial=True,
+            label=_("Exclude Course Staff"),
+            help_text=_(
+                "If selected, grades of course staff (participations who "
+                "have '%(permission)s' permissions) will be excluded."
+                % {"permission": pperm.assign_grade}),
```
inducer: Why not base this on the existing `pperm.included_in_grade_statistics`?

course/grades.py
inducer: (I'd much prefer a function, for example.)

course/grades.py
dzhuang: That's fair. I think then we should notice use if `False`, it will result in `NONE` in the csv.

course/grades.py
dzhuang: That's fair.

course/grades.py
dzhuang: In fact, the `csv_setting` object created here is intended to be an app global objects, since I am creating the object with its properties to be [`cached_properties`](https://docs.djangoproject.com/en/2.0/ref/utils/#django.utils.functional.cached_property), each of which need to be calculated only once.

However, I agree it's not elegant to use it here in this way. What about initialize the `csv_settings` object, i.e.,

```python
csv_settings = RelateCSVSettingsInitializer()
```
in `utils.py`, and then import it in `grades.py`. Will that be better?

dzhuang:
I felt this PR should be WIP before #461 is resolved, so that we can full test grade status and gradebook behavior.

PR #459, opened 2018-02-13
Added FAQ entry for changing things past deadlines

EnacheB:
There are many questions for how to do common things in RELATE from new TAs, I figured we should add this to the documentation.

### DIFF
--- a/doc/faq.rst
+++ b/doc/faq.rst
@@ -1,11 +1,6 @@
 Frequently Asked Questions
 ==========================
 
-What does the 'view' permission do?
-------------------------------------
-
-If you have it (the permission), you can see the pages in the flow. If
-you don't have it, you can't.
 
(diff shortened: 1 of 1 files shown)


inducer:
Thanks! This is *almost* correct, in that it really depends on the rule set that a course. So it'd be good if you could add a "typically" to your sentence, along with that caveat.

I think we could also stand to section the FAQ into "Administration" (running/setting up an instance), "Content Creation", and "Course Operations".

EnacheB:
Sounds good, but what do you mean by "rule set"?

inducer:
https://documen.tician.de/relate/flow.html#flow-rules

inducer:
Thanks!

inducer:
(Docs should update automatically in a bit.)

Issue #435, opened 2018-01-23
Switch back to un-forked bleach


inducer:
Based on guidance from https://github.com/mozilla/bleach/issues/348

willkg:
Feel free to ping me when you have a pull request and I can help review it.

inducer:
@willkg Thanks for offering to review!

dzhuang:
This issue was accidentally close by merging #477 . The changes to bleach part are reverted (thus we are still using forked version of bleach), with failed tests commented.

inducer:
This picks up a new urgency as the current setup appears to be occasionally crashy.

```
Traceback:

File "/home/andreas/my-relate-env3/lib/python3.6/site-packages/django/core/handlers/exception.py" in
inner
  35.             response = get_response(request)

File "/home/andreas/my-relate-env3/lib/python3.6/site-packages/django/core/handlers/base.py" in
_get_response
  128.                 response = self.process_exception_by_middleware(e, request)

File "/home/andreas/my-relate-env3/lib/python3.6/site-packages/django/core/handlers/base.py" in
_get_response
  126.                 response = wrapped_callback(request, *callback_args, **callback_kwargs)

File "./course/utils.py" in wrapper
  792.             response = f(pctx, *args, **kwargs)

File "./course/sandbox.py" in view_page_sandbox
  384.                             grade_data=None)

File "./course/page/code.py" in grade
  830.                             for fb_item in response.feedback))

File "./course/page/code.py" in <genexpr>
  830.                             for fb_item in response.feedback))

File "./course/page/code.py" in sanitize
  822.                 return bleach.clean(s, tags=["p", "pre"])

File "/home/andreas/my-relate-env3/lib/python3.6/site-packages/bleach/__init__.py" in clean
  99.     return cleaner.clean(text)

File "/home/andreas/my-relate-env3/lib/python3.6/site-packages/bleach/sanitizer.py" in clean
  210.             raise TypeError(message)

Exception Type: TypeError at /course/cs101-fa18/sandbox/page/
Exception Value: argument cannot be of 'list' type, must be of text type
Request information:
USER: davis68@illinois.edu

GET: No GET data

POST:
csrfmiddlewaretoken = 'vhtBjnSdd9vAMa6goFpBFHJMGlHnI7uL4iudaZ3eBidC2FeLJfiLFHML7CjpnFSo'
[...]

PR #465, opened 2018-02-18
Fully tested code.py, along with fixes.

dzhuang:


course/page/code.py
```diff
@@ -613,8 +613,7 @@ def transfer_attr(name):
         transfer_attr("names_for_user")
         transfer_attr("names_from_user")
 
-        if hasattr(self.page_desc, "test_code"):
-            run_req["test_code"] = self.get_test_code()
+        run_req["test_code"] = self.get_test_code()
```
dzhuang: `if hasattr(self.page_desc, "test_code"):` is done in `self.get_test_code()`.

course/page/code.py
```diff
@@ -908,18 +907,20 @@ def filter_img_attributes(tag, name, value):
...
                     elif name == "src":
-                        return is_allowed_data_uri([
-                            "image/png",
-                            "image/jpeg",
-                            ], value)
+                        if is_allowed_data_uri([
+                                "image/png",
+                                "image/jpeg",
+                                ], value):
+                            return bleach.sanitizer.VALUE_SAFE
```
dzhuang: Fix all  sanitized.

course/page/code.py
```diff
@@ -1136,7 +1138,8 @@ def grade(self, page_context, page_data, answer_data, grade_data):
 
         human_feedback_points = None
         if grade_data is not None:
-            if grade_data["feedback_text"] is not None:
+            assert grade_data["feedback_text"] is not None
```
dzhuang: `feedback_text` will never be `None`, if null, `feedback_text = ""`.

dzhuang:
@inducer Can you confirm the fixes commented above is valid? If yes, `code.py` get fully tested.

course/page/code.py
inducer: We should migrate off of the bleach fork I created. This would be a good opportunity to do so. See #435.

This change looks good otherwise.

course/page/code.py
inducer: LGTM

course/page/code.py
dzhuang: OIC, can you push some change for that?

course/page/code.py
dzhuang: OIC, can you push some change for that?

dzhuang: Tested, adding `protocols=["data"]` will do the work. I'll also make sure `<script src="data:..."></script>` will be sanitized.

inducer:
Argh. I'm not sure 23c8336 will do the job. I also concurrently wrote a substantially more complex version: 270db547. :(

dzhuang:
NVM, we have tests now.

dzhuang:
Sorry, the just merged #455 had brought some conflicts. I had to resolve that.  I think it's done now.

dzhuang:
Shall I add `Shall I add `bleach.ALLOWED_PROTOCOLS`  to `protocols`?  And Shall I refactor the bleach function out like 270db547de3a8cc488db939b6c7f889d7f56ffdb

dzhuang:
I think this PR will do the job (see patch and tests in https://github.com/mozilla/bleach/pull/350/). And it will be easier for us if that PR is merged.

dzhuang:
@inducer Can this be merged? I found I refactored some test function in this patch (my bad). If not merged, there will be conflicts when I submit other patches.

inducer:
Pushed some updates and a new test. The new (unit) test contradicts the results of the full-sandbox integration test.

On the whole, I am not at all convinced that this provides adequate defense against XSS. (which would be trivial to pull off for a motivated student and potentially disastrous)

dzhuang:
>  The new (unit) test contradicts the results of the full-sandbox integration test.

Sorry, the conflicts is on `assertResponseContextAnswerFeedbackNotContainsFeedback`, which failed to include `AnswerFeedback.bulk_feedback` as the haystack when doing the `assertIn` check. Fortunately, it only affected the failed tests above.

course/page/code.py
```diff
@@ -292,6 +294,97 @@ def request_python_run_with_retries(run_req, run_timeout, image=None, retry_coun
...
+        elif name == "src":
+            if is_allowed_data_uri([
+                    "image/png",
+                    "image/jpeg",
+                    ], value):
+                return True
+        else:
+            return False
+
+    else:
```
dzhuang: IMO, the logic below is redudant. After proceeding the logic below, bleach will do the same logic again. IOW, We are still [here](https://github.com/mozilla/bleach/blob/8706b5373b633407e804b1b2975edf8760067ff7/bleach/sanitizer.py#L508) after finishing the whole function.

The problem is, we didn't filter those built-in allowed tags with protocol allowed, in the logic above. So `<style src="data:,Evilcss">` just get passed.

course/page/code.py
```diff
@@ -292,6 +294,97 @@ def request_python_run_with_retries(run_req, run_timeout, image=None, retry_coun
...
+            # Remove replacement characters from unescaped characters
+            val_unescaped = val_unescaped.replace("\ufffd", "")
+
+            # Drop attributes with uri values that have
+            # protocols that aren't allowed
+            if (re.match(r'^[a-z0-9][-+.a-z0-9]*:', val_unescaped)
+                    and
+                    (val_unescaped.split(':')[0]
+                        not in bleach.ALLOWED_PROTOCOLS)):
+                return False
```
dzhuang: My comment above say, the logic here is redudant. The reason is it has no difference with the bleach logic, just repeating. That's not what you've expected. So we need to make it different. 

Let me guess: We are allowing only 'img', 'video' and 'source' to have uri attribute? If that's true, it's quite simple then, just fail those which have uri attribute is ok.

dzhuang: I simple strip `tags` which is not allowed, I think that will meet your requirements.

dzhuang:
Added another push. See the test result [here](https://ci.appveyor.com/project/inducer/relate/build/1.0.486#L441).

course/page/code.py
```diff
@@ -339,34 +338,7 @@ def filter_from_code_html_attributes(tag, name, value):
...
-            # Drop attributes with uri values that have
-            # protocols that aren't allowed
-            if (re.match(r'^[a-z0-9][-+.a-z0-9]*:', val_unescaped)
-                    and
-                    (val_unescaped.split(':')[0]
-                        not in bleach.ALLOWED_PROTOCOLS)):
-                return False
-
-        return name in bleach.ALLOWED_ATTRIBUTES
+    return False
```
inducer: Doesn't this mean that we delete attributes that `bleach` would normally allow?

course/page/code.py
```diff
@@ -378,6 +350,9 @@ def sanitize_from_code_html(s):
                                         "img"],
             attributes=filter_from_code_html_attributes,
 
+            # strip unwanted tags
+            strip=True,
+
```
inducer: As opposed to escaping them? Why?

tests/test_pages/test_code.py
```diff
@@ -860,8 +862,8 @@ def test_html_img(self):
         '<a src="data:text/html;base64,%s"</a>' % evil_b64_data,
         '<img src="https://Evil.com">',
```
inducer: Sorry, this is a debug leftover in a way, or at least it's a non-`data` URL in a list of data URLs, which is weird.

inducer:
In order to make progress on this, I think we should develop a sense of what's OK and what isn't.

* HTTP requests to URLs under user control? (say, from `` tags) (probably not OK)
* `data` URLs with JS mimetypes (anywhere, even in non-script settings) (IMO not OK, even if not harmful)
* ...

course/page/code.py
dzhuang: No.

course/page/code.py
dzhuang: https://github.com/mozilla/bleach/issues/348#issuecomment-367337250

course/page/code.py
dzhuang: I think it acts like removing items from blacklist, not a whitelist.

tests/test_pages/test_code.py
dzhuang: nvm.

course/page/code.py
dzhuang: In terms of attribute, I might be wrong, I'll check that.

course/page/code.py
dzhuang: Do you think `&lt;style src="data:,Evilcss"&gt;&lt;/style&gt;` still malicious?

course/page/code.py
dzhuang: Whether keeping those snippet or not, `bleach`'s default allowed tag-attributes are not allowed. Because that had been overriden by the callback `filter_from_code_html_attributes`.

`bleach`'s default allowed attributes are at [`here`](https://github.com/mozilla/bleach/blob/master/bleach/sanitizer.py#L51).

dzhuang:
And what about this@inducer? If the bleach is not ok right now. I'd rather comment those tests and revert the bleach part back, and merge other changes. IMO, I think the current bleach is not better than what it is now. 

For the failed test, it seems you mind the string url which contain 'evil'. But it's not possible to remove them by escape, as commented [here](https://github.com/mozilla/bleach/issues/348#issuecomment-367337250).

dzhuang:
This PR is accidently merged by merging #477. The bleach part is reverted (and tests commented) in b75a56daedd2d637029676ccf5ef21c01a97c6d7

Issue #430, opened 2018-01-18
Bug: reopen session won't change overall grade


dzhuang:
If a flow_session is graded with strategy `use_latest`, when the latest session is repopen, ~~the overall grade is expected to be using the grade of the second latest~~ the overall grade should be `None` (neither the grade before the flow_session is reopened, nor the grade of the seconde latest). However, the result is not changed (still display the previous grade of the repopened session).

inducer:
To handle this properly, "reopen session" would need to add a grade change for that `attemp_id` to "Unavailable".

That said, I'm not sure I totally find that behavior desirable.

inducer:
You've plowed ahead here in #423, but I'm still not sure what the right semantics are. Why not leave the latest grade in place until it's superseded?

dzhuang:
I think the gradebook is a main indicator for students and graders whether the he has finished/grade the flow. If we leave the latest grade as it is, sometimes students are encouraged not resubmit the session, because they might feel that new submission won't result in better grades. And for grader, he will not know there's a new session expecting regrades, when.the reopened session has been resubmitted.

inducer:
Well, I suppose all this is temporary anyway--before grades "count", I make sure that all open sessions are closed out, so it's just about what students see while that's going on. And I suppose removing the old grade might be less confusing. You've convinced me. :)

dzhuang:
Also, I think there's also another case, when `use_latest`, and the latest completed session is deleted (say) from admin. Then the grade change won't be able to react on that. That the same for `use_earliest` when the first completed session is reopened or deleted.

inducer:
I'd call inconsistent modifications in the admin "out of scope". The flow grades/grade book data keeping is redundant ("non-normalized"), largely on purpose. That means it relies on the code to keep those two views in sync. I think that any attempt to re-normalize this will just end up with a tangled mess, mostly because not every grade is associated with a flow.

dzhuang:
I just re-read through your previous comments, about the habit for doing grade "count". 

For me, close out all in-progress session will result in many 0-point assignments, you know, most of my pages were what we called variant pages, which generate different problem when a new session is opened (for the same gopp). It's common for my student to accidentally (or intensionally) open another new session after submission, not for the purpose for doing an extra homework.

dzhuang:
> I think that any attempt to re-normalize this will just end up with a tangled mess.

I agree. It's clever to generate the grades via a state machine.

PR #466, opened 2018-02-20
[WIP] Fix #263 #417 and #430.

### LINKED ISSUE #263: Bug: wrong overall grade when grading criterion changed between multiple submissions

dzhuang:
#263 #417: Bulk regrade resulted in wrong gradechange state (percentage) when `use_latest` 
#430: Reopening latest session if `use_latest` should result in `NONE`, instead of the grade of that session.

- [ ] Naming of newly introduced field `effective_time` in `GradeChange`
- [ ] In terms of backward compatibility, is the new migration right?
- [ ] More tests to cover changes

cc @inducer


dzhuang:
@inducer Can you review this? Thanks. The two TODOs need your verfication, thanks.

course/grades.py
```diff
@@ -118,9 +119,6 @@ def view_participant_grades(pctx, participation_id=None):
             if not (opp.shown_in_grade_book
                     and opp.shown_in_participant_grade_book):
                 continue
-        else:
-            if not opp.shown_in_grade_book:
-                continue
```
dzhuang: This is redundant as it had been done at this [line](https://github.com/inducer/relate/pull/466/files#diff-63d070b8acfd5aa88cf07944caf3f32cL104)

dzhuang:
@inducer Can you look at this PR and make sure the migration is correct? I wish to integrate this into my instance soon, new semester is coming. Thanks!

course/flow.py
```diff
@@ -1189,6 +1195,7 @@ def reopen_session(
         session,  # type: FlowSession
         force=False,  # type: bool
         suppress_log=False,  # type: bool
+        generate_gradechange=True,  # type: bool
```
inducer: Would prefer `generate_grade_change`.

course/grades.py
inducer: OK

course/grades.py
```diff
@@ -108,6 +108,7 @@ def view_participant_grades(pctx, participation_id=None):
                 "grade_time")
             .select_related("participation")
             .select_related("participation__user")
+            .select_related("flow_session")
```
inducer: Are these `select_related` still needed, given that we won't be looking at the flow sesions in grade finding?

course/models.py
```diff
@@ -1704,15 +1719,55 @@ def consume(self, iterable, set_is_superseded=False):
...
+                        key=lambda x: (x.grade_time, x.pk))
+
+                    _valid_percentages = (
+                        _valid_percentages_copy +
+                        [gchange.percentage() for gchange in _gchanges])
+
+                    if strategy == grade_aggregation_strategy.use_earliest:
+                        self.earliest_percentage = _valid_percentages[0]
+                    else:
+                        self.latest_percentage = _valid_percentages[-1]
```
inducer: Why is all this still needed?

course/models.py
```diff
@@ -1729,16 +1784,25 @@ def percentage(self):
...
-        elif strategy == grade_aggregation_strategy.min_grade:
-            return min(self.valid_percentages)
-        elif strategy == grade_aggregation_strategy.avg_grade:
-            return sum(self.valid_percentages)/len(self.valid_percentages)
+        if strategy in [grade_aggregation_strategy.max_grade,
+                        grade_aggregation_strategy.min_grade,
+                        grade_aggregation_strategy.avg_grade]:
+            numeric_percentages = [p for p in self.valid_percentages
+                                   if p is not None]
+            if len(numeric_percentages):
```
inducer: `if numeric_percentages:`

course/receivers.py
```diff
@@ -109,4 +109,40 @@ def may_preapprove_role(course, user):
...
+    last_gchange.pk = None
+    last_gchange.points = None
+    last_gchange.flow_session = None
+    last_gchange.creator = None
+    last_gchange.comment = None
+
+    from django.utils.timezone import now
+    last_gchange.grade_time = now()
+    last_gchange.state = grade_state_change_types.exempt
+    last_gchange.save()
```
inducer: "Exempt" is intended to mean that a student does not need to complete whatever is required for that grading opportunity, or at least that's the intention. (Yeah, I know--not well-documented.) Why is that the correct thing to do here?

inducer: (as opposed to a do-over, for example)

course/flow.py
dzhuang: Ok.

course/models.py
dzhuang: Ok.

course/receivers.py
dzhuang: > "Exempt" is intended to mean that a student does not need to complete whatever is required for that grading opportunity,

Sorry, I misunderstand that.

My idea is: since the session was deleted,
1. existing grade changes of that session should not be counted into the overall grade, thus we need to clear all grade changes related.
1. there should be no more NEW related grade changes for that session (or we throw an error),

course/grades.py
dzhuang: Yes, you're right. Will revert that.

course/flow.py
```diff
@@ -1139,12 +1139,15 @@ def grade_flow_session(
         # creator left as NULL
         gchange.flow_session = flow_session
         gchange.comment = comment
+        gchange.effective_time = (
+                # If not activity, use completion_time as effective_time
+                flow_session.last_activity() or flow_session.completion_time)
```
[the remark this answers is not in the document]
dzhuang: That's fair.

Your comment reminded me that, there might be some inconsistency (whether we respect the setting or not) is this case: the instructor modified the setting in the repo after the sessions had been graded,

course/grades.py
dzhuang: However, that is needed at least for `view_grades_by_opportunity`, see [here](https://github.com/inducer/relate/blob/f741e6b87c578c0ba2271ca764fdcfb4c417a2d3/course/templates/course/gradebook-by-opp.html#L171-L210).

course/models.py
dzhuang: I must admit the logic is not clear enough.

We still have the trouble after introducing `effective_time`:
Now only session-related grade_changes were using that attribute (For non-session-related ones, it's null). This is done intended. For comparing two gchange objects:
1. If either one is not session-related, use `grade_time` as key, then pk.
1. If both are session-related, use `effective_time`, then pk.

We can't have the above situation both use `effective_time`, because once a grade has been imported, the regrades previous submitted session will be overriden.

I'll try to simply the logic above.

course/receivers.py
dzhuang: Inconsistency were observed for a610287d30a85e79be051bbc757099579cd260e6.

course/models.py
dzhuang: Inconsistency were observed for a610287d30a85e79be051bbc757099579cd260e6, with order using using `__eq__ `, `__ge__`, `__le__` in `GradeChange` model.

dzhuang: See a610287d30a85e79be051bbc757099579cd260e6, tests [passed](https://ci.appveyor.com/project/inducer/relate/build/1.0.517) and [failed](https://ci.appveyor.com/project/dzhuang/relate/build/1.0.857). The [failed test](https://github.com/inducer/relate/blob/a610287d30a85e79be051bbc757099579cd260e6/tests/test_grades/test_grade_book.py#L469-L503) is actually the one you suggested in #423:

>> Can you give an example?

> I'll try:

> A: grade time 2015
> B: grade time 2016, flow completed 2014
> C: grade time 2017
> D: grade time 2018, flow completed 2013
> Then A < B < C

> But also D < B and C < D, which is a direct contradiction.

dzhuang: Updated. Removed `self.earliest_percentage` and `self.latest_percentage`

course/receivers.py
dzhuang: For this intentsion:

> "Exempt" is intended to mean that a student does not need to complete whatever is required for that grading opportunity, or at least that's the intention. 

P.S., I felt the currently it's not working this way, when new grades are to the opportunity, it throw [errors](https://github.com/inducer/relate/blob/816f832f87f2568df422593d3fd8a8541ac7f795/course/models.py#L1653-L1656).

dzhuang:
For record:

This might result in inconsistency when grade changes are ordered improperly.

> grades accumulation strategy: `use_latest`
> A: grade time 2015
> B: grade time 2016, flow completed 2014
> C: grade time 2017
> D: grade time 2018, flow completed 2013
> Then A < B < C

> But also D < B and C < D, which is a direct contradiction.

D's grade percentage should be the overall grade.

course/flow.py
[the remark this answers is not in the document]
dzhuang: Updated.

course/receivers.py
inducer: > currently, when new grades are append to the opportunity which already had a exempt gchange , it throw errors.

No, that's expected. "Exempt" is intended to mean, e.g. a student was sick and does no longer need to complete the assignment. As such, it is not expected that they complete the assignment, and it is considered a problem if they do. (because then the "exemption" was meaningless) Another plausible behavior would be to ignore any grade changes after "exempt" hits.

dzhuang:
@inducer What about this? I wish to put the `exempt` issue, since it's only debug-related.

inducer:
This is complicated (and terrifying) to review because of the long-term consequences. I just didn't have the time today. I'll try hard to get to it tomorrow night. Sorry!

dzhuang:
NVM and Thanks alot.

dzhuang:
Sorry for pushing you here @inducer. It seems I'll have to merge this into my instance tomorrow, if no progress is made here.

dzhuang:
In cases when reopening a session, https://github.com/inducer/relate/blob/855801d463120e0fc74de08466e5d4ac39bac46d/course/flow.py#L510 and https://github.com/inducer/relate/blob/855801d463120e0fc74de08466e5d4ac39bac46d/course/flow.py#L567 should also be adjusted.

dzhuang:
@inducer I've updated the code, looking forward to your review some day (not urging you to do it now, but hope that it will be merged someday).

PR #471, opened 2018-02-23
Fix up, add tests for 'relate' command line tool

inducer:
@dzhuang FYI: I noticed that this had been broken for a while. It's really my fault for not having added a test sooner.


dzhuang:
I seldom use that, no objection.

inducer:
Thanks for taking a look!

PR #472, opened 2018-02-23
Removed lxml dependency for ipynb utility.

### LINKED ISSUE #470: lxml dependency, HTML parse during markup render
Hi @dzhuang, I just spotted that 

* we've picked up a dependency on lxml, and
* each markup render involves a full HTML parse and re-stringify, just to get rid of some `<pre>` tags.

I'm not a fan of either of these two points, and I would like to see them fixed. Markup rendering is already way too slow. And lxml is not lightweight at all, requiring libxml2 and libxslt2. I've tried hard to keep Relate's dependencies to Python-only. On top of that, lxml is a bad tool for this job, with non-compliant HTML5 parsing. (`html5lib` exists for a reason)

dzhuang:
Fixed #470 .


dzhuang:
cc @inducer . Will that do the job? Thanks.

PR #473, opened 2018-02-24
Generate deprecation warning for custom page types

inducer:
cc @dzhuang 

Just to get the ball rolling on this. If you'd like extra per-site logic, feel free to add it.


inducer:
I guess I'll merge this, and we can do something finer-grained later.

PR #474, opened 2018-02-25
Add site-specific settings(warning/error) for deprecated custom pages.

dzhuang:


dzhuang:
cc @inducer What do you think about this PR?

inducer:
Seems fine. Some questions:

* I wonder whether we should really publicly document this. (I'd rather not.)
* How about allowing enforcement of earlier deadlines but "at the lastest 2019"? I'd like to avoid having this drag out.
* Is anyone on your end making heavy use of these things?

dzhuang:
> I wonder whether we should really publicly document this. (I'd rather not.)
 
Agreed.

> How about allowing enforcement of earlier deadlines but "at the lastest 2019"? I'd like to avoid having this drag out.

Agreed.

> Is anyone on your end making heavy use of these things?

Me only. But I can migrate those pages without problem.

dzhuang:
I'll remove that from docs, and update the settings.

dzhuang:
Updated.

dzhuang:
Also, should this be removed?
> Instructors can implement custom question/page types in Python.

inducer:
> Also, should this be removed?

Yes, definitely.

inducer:
LGTM aside from the doc/README removal.

dzhuang:
>> Also, should this be removed?

> Yes, definitely.

Done. And I think it's ok to merge.

PR #484, opened 2018-03-08
Fix have_valid_page context in page sandbox (0ea0802 continued).

dzhuang:


dzhuang:
@inducer Something's wrong with your Gitlab instance. 

curl https://gitlab.tiker.net/inducer/ci-support/raw/master/prepare-and-run-flake8.sh failed.

PR #487, opened 2018-03-11
Make URLs relative-friendly

### LINKED ISSUE #486: Login/logout redirects to root, not home route
Code like [this](https://github.com/inducer/relate/blob/4b405525b9d9b5859f671ab3ece897e1555952a5/course/auth.py#L414-L431) assumes that the application is mounted at domain root, while in my case it isn't.

Before redirecting, we should check if it's empty and in the case, set it to the route `relate-home`.

ishitatsuyuki:
Fix #486


inducer:
Also, thank you for your contribution!

dzhuang:
LGTM. Thanks.

PR #488, opened 2018-03-12
Migrate all dependencies to NPM

ishitatsuyuki:
NPM contains the up-to-date version of the dependencies, and uses semantic versioning to minimize upgrade breakage. It's also much faster than bower, and actively maintained.

There were some changes during the migration of `datatables-i18n` (unmaintained) to the official plugins repository.


dzhuang:
Thank you for the contribution. What do you think of this PR as compared to #412 ? cc @inducer

ishitatsuyuki:
Well, I think incrementally migrating is the best solution.

This PR only contains minimal change, and I think we should have this merged while we continue the work on migrating to webpack.

inducer:
@ishitatsuyuki Thank you for submitting this!

#412 has the advantage of using webpack, which might give us faster page loads. On the other hand, this is ready now (and the sooner we are off bower, the better), whereas #412 has a variety of unadressed to-do items.

inducer:
i would prefer choosing exactly one package manager (between yarn or npm), and I'd personally prefer yarn over npm. If we allow both, `yarn.lock` and `packages.json` might fall out of sync.

Aside from that, I'd be happy to see this merged.

ishitatsuyuki:
It... doesn't. package.json specifies the semantic (relaxed) version requirements, while yarn.lock saves the resolved dependency tree with exact versions.

(There's also package-lock.json introduced in NPM 5 which does duplicate yarn's functionality, but that's not widespread yet)

inducer:
You're right, sorry. D'oh.

course/templatetags/coursetags.py
```diff
@@ -36,8 +36,8 @@ def __init__(self, variable):
         self.variable = variable
 
     def render(self, context):
-        js_lang_name = to_js_lang_name(translation.get_language())
-        context[self.variable] = js_lang_name
+        datatables_lang_name = to_datatables_lang_name(translation.get_language())
+        context[self.variable] = datatables_lang_name
```
dzhuang: Since this is refactored, shall we also rename the following `get_current_js_lang_name` to explicitly indicate it's for datatable js only?

inducer:
LGTM. @dzhuang?

dzhuang:
I'll add tests for uncovered changes. And since the template tag is for datatables only, we might better rename it.

PR #496, opened 2018-03-13
Make the grading buttons sane

ishitatsuyuki:
- Implement proper POST-redirect instead of opening many pages
- Remove the functionality of top submit buttons
- Fixed some bad/deprecated HTML tags

### DIFF
--- a/course/grading.py
+++ b/course/grading.py
@@ -313,6 +313,13 @@ def grade_flow_page(pctx, flow_session_id, page_ordinal):
 
                 prev_grade_id = _save_grade(fpctx, flow_session, most_recent_grade,
                         bulk_feedback_json, now_datetime)
+
+                if "submit_next_page" in pctx.request.POST:
+                    return redirect("relate-grade_flow_page",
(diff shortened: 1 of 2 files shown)


inducer:
Thanks for your contribution! The behavior you removed, unfortunately, is an intentional (if ugly) workaround for the fact that submitting a grade for a code page will reexecute the code, which may take on the order of 20 seconds. The current behavior has the advantage of letting the grader continue their work while the current page submits in the background.

There would be many less ugly ways of dealing with the same problem:

* Submit the grade computation to the (Celery) job queue so as to not stall the user.
* Make the grading interface dynamic/AJAXy, so that, say an XHR for grade submission and the fetch of the next page can occur concurrently. In general, I would view this as the way forward, since it permits even more improvements: For instance, we could even prefetch the next/previous pages to make that transition more seamless. https://www.gradescope.com/ is a good example that a good grading interface on its own is worth something.

ishitatsuyuki:
I see the point here. It's quite hard to get AJAX done right, so probably I will do only the Celery implementation in this PR.

PR #499, opened 2018-03-15
Improve rendering of page title

dzhuang:
- [x]  add tests

### DIFF
--- a/course/page/base.py
+++ b/course/page/base.py
@@ -25,6 +25,7 @@ THE SOFTWARE.
 """
 
 import six
+import re
 
 import django.forms as forms
 
@@ -388,8 +389,9 @@ class PageBase(object):
             self.is_optional_page = getattr(page_desc, "is_optional_page", False)
 
         else:
-            from warnings import warn
-            warn(_("Not passing page_desc to PageBase.__init__ is deprecated"),
+            if vctx is not None:
+                vctx.add_warning(
+                    _("Not passing page_desc to PageBase.__init__ is deprecated"),
                     DeprecationWarning)
             id = page_desc
             del page_desc
@@ -765,9 +767,11 @@ class PageBaseWithTitle(PageBase):
             try:
                 md_body = self.markup_body_for_title()
             except NotImplementedError:
-                from warnings import warn
-                warn(_("PageBaseWithTitle subclass '%s' does not implement "
-                        "markdown_body_for_title()")
+                if vctx is not None:
+                    vctx.add_warning(
+                        location,
+                        _("PageBaseWithTitle subclass '%s' does not implement "
+                          "markdown_body_for_title()")
                         % type(self).__name__)
             else:
                 from course.content import extract_title_from_markup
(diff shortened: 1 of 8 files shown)


dzhuang:
@inducer What do you think of this?

course/page/base.py
```diff
@@ -388,8 +389,9 @@ def __init__(self, vctx, location, page_desc):
             self.is_optional_page = getattr(page_desc, "is_optional_page", False)
 
         else:
-            from warnings import warn
-            warn(_("Not passing page_desc to PageBase.__init__ is deprecated"),
+            if vctx is not None:
+                vctx.add_warning(
+                    _("Not passing page_desc to PageBase.__init__ is deprecated"),
```
inducer: This doesn't have much to do with course content--I don't think this warning belongs in validation output.

course/page/base.py
```diff
@@ -765,9 +767,11 @@ def __init__(self, vctx, location, page_desc):
...
                 md_body = self.markup_body_for_title()
             except NotImplementedError:
-                from warnings import warn
-                warn(_("PageBaseWithTitle subclass '%s' does not implement "
-                        "markdown_body_for_title()")
+                if vctx is not None:
+                    vctx.add_warning(
+                        location,
+                        _("PageBaseWithTitle subclass '%s' does not implement "
+                          "markdown_body_for_title()")
```
inducer: Same.

inducer: Also, that warning gets the method name wrong. (Not introduced by you, but worth fixing nonetheless.)

course/page/base.py
```diff
@@ -780,6 +784,13 @@ def __init__(self, vctx, location, page_desc):
                         _("no title found in body or title attribute"))
                     % (location))
 
+        from markdown import markdown
+        from django.utils.html import strip_tags
+        title = strip_tags(markdown(title))
```
inducer: Assuming `strip_tags` removes all HTML? If so, that seems fine. The title *is* fished out of the Markdown after all... Though I still don't fully understand the failure mode, as I noted.

course/page/base.py
dzhuang: You're right, Actually I'm now reverting that.

course/page/base.py
dzhuang: Both `**important**` and  `<b>important</b>`  are expected to be `important` in title.

course/page/base.py
dzhuang: > that warning gets the method name wrong

Will fix that.

dzhuang:
Updated.

course/templates/course/flow-page.html
```diff
@@ -285,7 +285,7 @@
               {% for other_page_data in all_page_data %}
               <li>
                 <a href="{% url "relate-view_flow_page" course.identifier flow_session.id other_page_data.page_ordinal %}">
-                  {{ other_page_data.human_readable_ordinal }}: {{ other_page_data.title }}
+                  {{ other_page_data.human_readable_ordinal }}: {{ other_page_data.title|safe|striptags }}
```
inducer: Why is it `|striptags` in some places and not in others?

dzhuang: Sorry, it's my bad, that should be removed.  I should have proof read before submit.

dzhuang: Updated.

inducer:
LGTM

dzhuang:
Thanks for reviewing this.

PR #516, opened 2018-03-28
More readable data type error messages

adamjstewart:
Because this:
```
'data points' does not have correct data type--got: 'float64', expected: 'int64'
```
is much more readable than this:
```
'data points' does not have correct data type--got: 'f', expected: 'i'
```
This is also inline with the error message for `check_numpy_array_sanity`:
```
'data points' does not consist of floating point numbers--got: 'int64'
```

### DIFF
--- a/course/page/code_feedback.py
+++ b/course/page/code_feedback.py
@@ -92,7 +92,7 @@ class Feedback:
             return bad(
                     "'%s' does not have correct data type--"
                     "got: '%s', expected: '%s'" % (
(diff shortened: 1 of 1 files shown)


inducer:
Thanks!

inducer:
(I'll rebuild the container image sometime soon.)

inducer:
Built, validated, and pushed a new image to the UIUC instance, your new message (and prior fixes should) be in there.

PR #519, opened 2018-03-29
Switch to amd64 docker image

inducer:


inducer:
FYI @dzhuang 

Built and pushed new images, too. Seems a bit quicker as amd64--but that may be subjective.

Not sure if you have any use for tensorflow, but there's now a version of the Docker image that includes TensorFlow, if you have a use for such a thing. (Someone here asked for it.)

dzhuang:
Thank you!

> Not sure if you have any use for tensorflow, but there's now a version of the Docker image that includes TensorFlow,

That's great!

PR #522, opened 2018-03-30
Added tests for versioning.py

dzhuang:
Also passed tests with dulwich == 0.18 (before changing the dependancy).

### DIFF
--- a/course/versioning.py
+++ b/course/versioning.py
@@ -75,12 +75,6 @@ if False:
 # }}}
 
 
-class AutoAcceptPolicy(paramiko.client.MissingHostKeyPolicy):
-    def missing_host_key(self, client, hostname, key):
-        # simply accept the key
-        return
-
-
 def _remove_prefix(prefix, s):
     # type: (bytes, bytes) -> bytes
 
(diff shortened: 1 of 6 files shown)


dzhuang:
This should work around #492. @ishitatsuyuki @inducer What do you think of this?

ishitatsuyuki:
Not a bad idea, but any reason for not using ParamikoSSHVendor?

dzhuang:
I guess @inducer 's idea was to use a customized `progress_stderr` callback, but there's no interface to do that.

ishitatsuyuki:
It looks like we can do that today (correct me if I'm wrong). Can we migrate?

dzhuang:
Of course. I also think the code here can be moved to dulwich as a PR.

dzhuang:
@ishitatsuyuki So do you think this is mergeable or you prefered to do something on the `dulwich` side (add that interface) ?

ishitatsuyuki:
I would prefer adding the tests to dulwich instead, and simply migrate to dulwich's `ParamikoSSHVendor`.

dzhuang:
Ok. However, I don't think my test (currenty unittest only) meet the requirement of `dulwcih`. Let's discuss at dulwich/dulwich#364.

dzhuang:
@inducer Do you think  it's ok to migrate to  dulwich's built-in ParamikoSSHVendor?

dzhuang:
https://github.com/dulwich/dulwich/pull/618

dzhuang:
This would fail for dulwich 0.18, I'll update the dependency in a while.

dzhuang:
Meged. Thanks for the suggestions @ishitatsuyuki . @inducer dulwich need to upgraded to 0.19 or above.

PR #535, opened 2018-04-12
Add test utils

dzhuang:
@inducer Can you review the 2nd commit?

### DIFF
--- a/course/content.py
+++ b/course/content.py
@@ -1510,7 +1510,7 @@ def instantiate_flow_page(location, repo, page_desc, commit_sha):
 # }}}
 
 
-class CourseCommitShaDoesNotExist(Exception):
+class CourseCommitSHADoesNotExist(Exception):
     pass
 
 
@@ -1525,13 +1525,13 @@ def get_course_commit_sha(course, participation, repo=None,
         if isinstance(repo, SubdirRepoWrapper):
             repo = repo.repo
         try:
-            repo[preview_sha.encode()]
+            repo[commit_sha.encode()]
         except KeyError:
             if raise_on_nonexistent_preview_commit:
-                raise CourseCommitShaDoesNotExist(
+                raise CourseCommitSHADoesNotExist(
                     _("Preview revision '%s' does not exist--"
                       "showing active course content instead."
-                      % participation.preview_git_commit_sha))
+                      % commit_sha))
             return False
 
         return True
--- a/course/content.py
+++ b/course/content.py
@@ -1514,29 +1514,40 @@ class CourseCommitShaDoesNotExist(Exception):
     pass
 
 
-def get_course_commit_sha(course, participation,
+def get_course_commit_sha(course, participation, repo=None,
                           raise_on_nonexistent_preview_commit=False):
-    # type: (Course, Optional[Participation], Optional[bool]) -> bytes
+    # type: (Course, Optional[Participation], Optional[Repo_ish], Optional[bool]) -> bytes  # noqa
 
     sha = course.active_git_commit_sha
+
+    def is_commit_sha_valid(repo, commit_sha):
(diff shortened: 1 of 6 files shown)


course/content.py
```diff
@@ -1528,7 +1530,14 @@ def get_course_commit_sha(course, participation):
...
                     repo[preview_sha.encode()]
                 except KeyError:
+                    if raise_on_error:
+                        raise CourseCommitShaDoesNotExist(
+                            _("Preview revision '%s' does not exist--"
+                              "showing active course content instead."
+                              % participation.preview_git_commit_sha))
                     preview_sha = None
+                finally:
+                    repo.close()
```
inducer: Is this necessary if the `repo` is inside a `with` block here?

course/utils.py
```diff
@@ -598,40 +598,16 @@ def __init__(self, request, course_identifier):
         from course.views import check_course_state
         check_course_state(self.course, self.participation)
 
-        self.course_commit_sha = get_course_commit_sha(
-                self.course, self.participation)
```
inducer: Ouch. That's kind of embarrassing. How did that get here? Double-duplication?

course/content.py
```diff
@@ -1510,10 +1510,12 @@ def instantiate_flow_page(location, repo, page_desc, commit_sha):
...
 
 
-def get_course_commit_sha(course, participation):
-    # type: (Course, Optional[Participation]) -> bytes
+class CourseCommitShaDoesNotExist(Exception):
+    pass
+
 
-    # logic duplicated in course.utils.CoursePageContext
+def get_course_commit_sha(course, participation, raise_on_error=False):
```
inducer: `raise_on_error` -> `raise_on_nonexistent_preview_commit`?

course/utils.py
```diff
@@ -598,40 +598,16 @@ def __init__(self, request, course_identifier):
...
-                if preview_sha is not None:
-                    sha = preview_sha
+        try:
+            sha = get_course_commit_sha(
+                self.course, self.participation, raise_on_error=True)
+        except CourseCommitShaDoesNotExist as e:
+            from django.contrib import messages
+            messages.add_message(request, messages.ERROR, str(e))
+
+            sha = self.course.active_git_commit_sha.encode()
```
inducer: A  minor point is that `get_course_commit_sha` opens another handle on the repo. I don't know what the performance impact of that is--but it can't be too bad, because we were already doing that...

course/content.py
dzhuang: You're right. Removed.

course/content.py
dzhuang: Updated.

course/utils.py
dzhuang: What about the latest push?

course/content.py
```diff
@@ -1510,25 +1510,44 @@ def instantiate_flow_page(location, repo, page_desc, commit_sha):
...
-def get_course_commit_sha(course, participation):
-    # type: (Course, Optional[Participation]) -> bytes
+class CourseCommitShaDoesNotExist(Exception):
+    pass
+
 
-    # logic duplicated in course.utils.CoursePageContext
+def get_course_commit_sha(course, participation, repo=None,
+                          raise_on_nonexistent_preview_commit=False):
+    # type: (Course, Optional[Participation], Optional[Repo_ish], Optional[bool]) -> bytes  # noqa
```
inducer: `Optional` is not the correct thing to use here. `Optional` indicates that a value may be `None`, not that it is an optional argument.

dzhuang: I knew that, but `repo` here can be `None`.

dzhuang: Er, the actual issue was with the inner function defined.

PR #549, opened 2018-04-18
Pin pysaml2<=4.4.0

dzhuang:
According to log in tests:

> djangosaml2 0.16.11 has requirement pysaml2==4.4.0, but you'll have pysaml2 4.5.0 which is incompatible.


inducer:
Noted, seems fine. Sorry for not getting to this sooner.

Issue #551, opened 2018-04-19
Discussion: about FlowRuleException


dzhuang:
Curently, there seems no way for stage 3 to know whether the exception will be granted to a new session, or an existing session created at stage2.  

The point is that, when the stage 3 form is valid, and a POST is done without creating a new access rule and a new grading rule, the message is always "Exception granted to XXX in course XXX", however, there's actually no `FlowRuleException` created in the process. 

If the above process is for an exising session, (that message is kind of misleading, thought I know this is a rare case). If the process is for a new session is created in stage 2, the message is correct, but there's no way to track which sessions are created throught this exception granting process, if the stage 3 is post (or even by passed) without creating new access/grading rules.

dzhuang:
I'll change the message if 3rd stage is posted with no exception created. However, how to track to exception granted, in terms of a new session,  is still an issue.

PR #552, opened 2018-04-20
Added tests for views, added more messages for exception creation

dzhuang:
1. Added warning if no FlowRuleException were actually created, #551 
1. Added message when a session is created when creating FlowRuleException at stage 2.
1. Added message when the access rule tag of a session is updated/removed during creating FlowRuleException.

### DIFF
--- a/course/views.py
+++ b/course/views.py
@@ -97,7 +97,7 @@ if False:
 # }}}
 
 
-NONE_SESSION_TAG = "<<<NONE>>>"  # noqa
+NONE_SESSION_TAG = string_concat("<<<", _("NONE"), ">>>")  # noqa
 
 
 # {{{ home
@@ -353,7 +353,7 @@ def get_repo_file_response(repo, path, commit_sha):
         raise http.Http404()
 
     from mimetypes import guess_type
-    content_type, _ = guess_type(path)
+    content_type, __ = guess_type(path)
 
     if content_type is None:
         content_type = "application/octet-stream"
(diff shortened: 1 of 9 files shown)


course/views.py
```diff
@@ -1164,16 +1185,27 @@ def grant_exception_stage_3(pctx, participation_id, flow_id, session_id):
...
                 session.save()
+                session_access_rules_tag_changed = True
+
+                if new_access_rules_tag is not None:
+                    msg = _("Access rules tag of the selected session "
+                            "updated to '%s'.") % new_access_rules_tag
+                else:
+                    msg = _("Removed access rules tag of the selected session.")
+
+                messages.add_message(pctx.request, messages.SUCCESS, msg)
```
inducer: Why doesn't this set `exception_created` to `True`?

course/views.py
```diff
@@ -1224,16 +1256,38 @@ def grant_exception_stage_3(pctx, participation_id, flow_id, session_id):
...
             # }}}
 
-            messages.add_message(pctx.request, messages.SUCCESS,
-                    ugettext(
-                        "Exception granted to '%(participation)s' "
-                        "for '%(flow_id)s'.")
-                    % {
-                        'participation': participation,
-                        'flow_id': flow_id})
+            if exception_created:
```
inducer: Wouldn't this be better as two independent messages--"grading exc created" and "access exc created"?

course/views.py
dzhuang: Because it doen't guarentee an exception object was or will be created. I've post a discussion (#551 ) about this.

course/views.py
dzhuang: Yes, that's better. I'll make the change.

dzhuang: Done by #572

PR #556, opened 2018-04-20
Fix repo file visit for folder depth more than 2, on Windows.

dzhuang:
This is because path sep is differernt from that of Linux ("\\" versus "/"). Without this, images put in 2nd depth folder won't be reachable on Windows, and Subfolder repo won't work either.

This is a continued fix for #550

### DIFF
--- a/course/content.py
+++ b/course/content.py
@@ -29,6 +29,7 @@ from typing import cast, Union, Text
 from django.conf import settings
 from django.utils.translation import ugettext as _
 
+import os
 import re
 import datetime
 import six
@@ -197,8 +198,7 @@ def get_true_repo_and_path(repo, path):
 def get_course_repo_path(course):
     # type: (Course) -> Text
 
-    from os.path import join
-    return join(settings.GIT_ROOT, course.identifier)
+    return os.path.join(settings.GIT_ROOT, course.identifier)
 
 
 def get_course_repo(course):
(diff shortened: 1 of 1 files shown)


course/content.py
```diff
@@ -224,8 +224,7 @@ def get_repo_blob(repo, full_name, commit_sha, allow_tree=True):
 
     dul_repo, full_name = get_true_repo_and_path(repo, full_name)
 
-    from os.path import split
-    names = split(full_name)
+    names = os.path.normpath(full_name).split(os.sep)
```
dzhuang: The fix is here.

dzhuang:
cc @inducer

course/content.py
inducer: Why is `str.split` better than `os.path.split`?

inducer:
I'm looking at this code now, and I think it has fundamentally the wrong idea in terms of supporting Windows. Path specification in Relate should consistently use `/`, otherwise you end up with Linux-compatible and Windows-compatible repos, containing their specific brand of path separator, but you can't use a "Windows" relate repo on a "Unix" Relate instance.

dzhuang:
Sorry for my late response. I need sometime to remember why I did that change and then what you mean.
I should have manually test the code instead of a unit test at that time.  I think it will help if you can provide the traceback in #767 and I will try to add a testcase to fix this.

inducer:
Thanks for getting back to me. Responded in #767.

course/content.py
dzhuang: `os.path.split` result in a 2-tuple, while is not appropriate where the depth of the file path is more than 2.

PR #564, opened 2018-04-22
Add assertion error msg for inconsistent test.

dzhuang:


dzhuang:
@inducer With this patch, there will be error msg next time it failed.

PR #584, opened 2018-05-24
Validate Event.kind

### LINKED ISSUE #583: BUG: Event object creation not validated
1. `kind` can be any string, although it declared in  `help_text`  that `Should be lower_case_with_underscores, no spaces allowed.`.
2.  if `end_time` is not None, it must not be a value ahead of `time` (start_time).

dzhuang:
Fix #583 .

### DIFF
--- a/course/models.py
+++ b/course/models.py
@@ -52,7 +52,7 @@ from course.constants import (  # noqa
         exam_ticket_states, EXAM_TICKET_STATE_CHOICES,
         participation_permission, PARTICIPATION_PERMISSION_CHOICES,
 
-        COURSE_ID_REGEX, GRADING_OPP_ID_REGEX, NAME_VALID_REGEX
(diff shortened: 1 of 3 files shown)


course/models.py
```diff
@@ -291,7 +291,14 @@ class Event(models.Model):
             # Translators: format of event kind in Event model
             help_text=_("Should be lower_case_with_underscores, no spaces "
             "allowed."),
-            verbose_name=_('Kind of event'))
+            verbose_name=_('Kind of event'),
+            validators=[
+                RegexValidator(
+                    "^" + EVENT_KIND_REGEX + "$",
+                    message=_("Should be lower_case_with_underscores, no spaces "
```
dzhuang: @inducer Can you take a look at this? The validation error message is exactly the same with the help_text of the field. Will that look weird? Thanks.

inducer: That's fine IMO.

dzhuang:
Then I'll merge this. Thanks.

PR #586, opened 2018-05-25
Render events via Ajax.

dzhuang:
Used AJAX to fetch the events and event_info, this made it possible to create/update/delete events in FullCalendar UI without refresh the page. 

1. Added listview (month) in FullCalendar UI.
2. FullCalendar.js need to be updated to 3.9.0.

### DIFF
--- a/course/calendar.py
+++ b/course/calendar.py
@@ -26,27 +26,29 @@ THE SOFTWARE.
 
 import six
 from six.moves import range
+import datetime
+from bootstrap3_datetime.widgets import DateTimePicker
+from crispy_forms.layout import Submit
 
 from django.utils.translation import (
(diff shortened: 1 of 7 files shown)


dzhuang:
cc @inducer

dzhuang:
According to https://stackoverflow.com/a/40832304/3437454, there's no way to prevent refetching events when using data feeds. To lower the burden of AJAX refetch, calendar for participations with no edit_events pperm uses JSON objects, while that for participations with edit_events pperms uses AJAX.

dzhuang:
I think I'd better base the PR on https://gitlab.tiker.net/inducer/relate/merge_requests/42 . So this PR is now closed.

Issue #590, opened 2018-07-05
Quotation is lack in logout process


proelbtn:
When I tried to logout, I caught the following error:

```
Uncaught SyntaxError: Invalid regular expression flags
```

I found the content of onclick of sign out button is a little strange. The url isn't quoted.

<img width="272" alt="screen shot 2018-07-05 at 21 45 40" src="https://user-images.githubusercontent.com/18246032/42323965-ce69dd9c-809c-11e8-9796-c6c174ebd34f.png">

so, I caught this error.

- https://github.com/inducer/relate/blob/master/course/auth.py#L941

PR #591, opened 2018-07-05
[FIX] Quotation is lack in logout process

proelbtn:
fixes: inducer/relate#590


inducer:
Thanks for contributing this fix!

PR #600, opened 2018-08-20
Fix validation check for duplicate (kind, no-ordinal) events

inducer:
cc @dzhuang


dzhuang:
Thanks. I'll fix the test in a while. P.S., What about #584 ?

inducer:
> Thanks. I'll fix the test in a while.

Thanks!

inducer:
> What about #584?

Sorry, lost track of it. Will take a look now.

inducer:
@dzhuang Codecov seems to have broken. Any clue what happened there?

inducer:
(Merging because I don't think the current set of changes has anything to do with codecov)

dzhuang:
Thanks!

PR #605, opened 2018-08-29
Fix failed creation of recurring events without ordinal for non-English langs

### LINKED ISSUE #603: Create recurring event: cannot leave ordinal unspecified twice
For example:

* Create 20 weekly events `office_andreas`, leaving the ordinal unspecified, from (say) a Wednesday at 3:30pm
* Create 20 weekly events `office_andreas`, again leaving the ordinal unspecified, from (say) a Friday at 3:30pm
* "An event with this kind, ordinal, and course already exists." (Paraphrasing from German.)

@dzhuang I distinctly recall that this worked at some point...

dzhuang:
Fixed #603.


dzhuang:
cc @inducer

PR #612, opened 2018-09-05
Fix case when email suffix match but need approval.

### LINKED ISSUE #608: Students are able to enroll without approval
https://github.com/inducer/relate/blob/b0e64c93799cb8b603880f13d5d2c07f4117d06a/course/enrollment.py#L251-L267

Looking at the code in the enrollment view, it looks like students are able to enroll without approval as long as their email suffix matches. This is causing some operational disruption on our end.

@dzhuang Since you were the last one to touch the code, could you explain?

cc @eecarrier

dzhuang:
Fixed #608: when no preapproval, participations with matched email suffix also need to be approved.

### DIFF
--- a/course/enrollment.py
+++ b/course/enrollment.py
@@ -240,6 +240,16 @@ def enroll_view(request, course_identifier):
         else:
             return email.endswith("@%s" % suffix) or email.endswith(".%s" % suffix)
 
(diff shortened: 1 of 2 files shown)


inducer:
Thanks!

PR #615, opened 2018-09-09
Make recurring-event tests tolerant to dailight savings time

inducer:
e.g. https://gitlab.tiker.net/inducer/relate/-/jobs/61483

cc @dzhuang


dzhuang:
Thanks!

PR #622, opened 2018-10-02
Only mark flow grade unavailble upon unavailable page grade if page c…

inducer:
…arries value

### DIFF
--- a/course/flow.py
+++ b/course/flow.py
@@ -853,7 +853,9 @@ def gather_grade_info(
 
             if feedback is None or feedback.correctness is None:
                 unknown_count += 1
-                points = None
+                if not grade.max_points:
+                    points = None
+
                 continue
 
             max_reachable_points += grade.max_points


dzhuang:
Sorry for being late.

The broken test cases are testing that, when a page had been previously graded, and then a grader removed the grade (by submitting the human grading form with points field empty), the grading status of the page is then "None". 

Your changes made that impossible, and grade those pages with 0 points instead.

inducer:
Good catch. My previous logic was flipped. How about now?

dzhuang:
I think that's ok now. I'm wondering when that will happen? (for writing tests).

dzhuang:
Then this issue is related to #623? If yes, there might be a problem, from the perspective of graders, how can a grader identify and find ungraded optional pages? Currently, my asistants identify those pages by looking for flows which had not generated grades.

inducer:
>  how can a grader identify and find ungraded optional pages?

I think that clicking "view page grades" in the "grading opportunity" view should show grades for all pages and all students.

PR #623, opened 2018-10-02
Count points from optional pages

### LINKED ISSUE #621: Optional pages don't contribute to the point count
@dzhuang When you introduced `is_optional_page: True`, you made it so that optional pages do not contribute to the point count. I'm not sure I understand the logic behind that. Could you explain?

inducer:
Closes #621 

@dzhuang Could you review/comment?

### DIFF
--- a/course/flow.py
+++ b/course/flow.py
@@ -841,6 +841,12 @@ def gather_grade_info(
                 optional_unknown_count += 1
                 continue
 
+            else:
+                page_points = grade.max_points*feedback.correctness
+
+                if points is not None:
+                    points += page_points
+
             if feedback.correctness == 1:
                 optional_fully_correct_count += 1
             elif feedback.correctness == 0:


dzhuang:
See discussion in #621. I think what you want is a bonus page?

inducer:
> I think what you want is a bonus page?

Yeah, pretty much. Do we have a dfiferent mechanism for those? Because optional + has value makes a lot of sense to me to express that.

dzhuang:
I agree. So there are some agreements need to be reached to achieve that:

1. Warning that there're bonus page unanswered (for pages that have non-zero values)
1. Display `bonus points` in those pages
1. Any more?

inducer:
> Warning that there're bonus page unanswered (for pages that have non-zero values), when a user is finishing a flow with bonus pages.

Only if they carry points.

> Display bonus points in those pages

Could simply be shown as "5 points (optional)".

We may want a separate grade sum for "bonus points" on the results page--maybe.

dzhuang:
That sounds reasonable, but I have no time to do that in a while. Maybe I can help in testing if you can submit a patch.

dzhuang:
So I'll wait for your ping :)

PR #626, opened 2018-10-22
Pass actual user source code (not a code object) to grading code in u…

inducer:
…ser_code


inducer:
Since github is a potato at the moment, using Gitlab instead: https://gitlab.tiker.net/inducer/relate/merge_requests/65

PR #628, opened 2018-10-30
Improve size and autofocus behavior of codemirror widgets

inducer:
@dzhuang What do you think? Any objections to this?


dzhuang:
I'm ok with that. Thanks.

PR #647, opened 2019-04-15
CAS support.

Mihara:
Greetings. We *(the [Center for Pedagogical Mastery](https://cpm.dogm.mos.ru/))* are currently evaluating RELATE as an alternative to [Open edX](https://open.edx.org/), *(that is, we are hoping to get rid of it as soon as possible, because it's really not a good fit for how we teach, and RELATE appears to be much closer to ideal for us)*. To do that, the first thing we needed to do is to add support for authorization over [CAS protocol](https://en.wikipedia.org/wiki/Central_Authentication_Service), keeping it extensible for the unorthodox uses we've put it for. *(We have a separate in-house system managing course subscriptions, which communicates which courses people should be subscribed to as CAS attributes, so the basic attribute mapper that django-cas-ng provides is insufficient.)* This, of course, should be generally useful.

To that end, I present this PR, hoping you could accept it. It works, and passes all the tests that were there already, but I couldn't think of what kind of tests would be appropriate to test the new functionality. Should you deem it unsuitable, I'm ready to hack it until it is.

### DIFF
--- a/cas-config/.gitignore
+++ b/cas-config/.gitignore
@@ -1,0 +1,1 @@
+callback.py
--- a/cas-config/models.py
+++ b/cas-config/models.py
@@ -1,0 +1,22 @@
+from django.conf import settings
+from django.dispatch import receiver
+from django_cas_ng.signals import cas_user_authenticated
+
+
+@receiver(cas_user_authenticated)
+def cas_attribute_callback_manager(sender, user, created, attributes, ticket, service, request, **kwargs):
+    """
+    Upon receiving a signal from django-cas-ng that a user was successfully logged in over CAS,
+    if a CAS_ATTRIBUTE_CALLBACK is defined in settings,
+    call it with the data we got from this signal.
+
+    This is a bit of a roundabout way to do it, but allows to flexibly configure it without
+    touching the main RELATE code at all.
+    """
+    callback = getattr(settings, 'CAS_ATTRIBUTE_CALLBACK', None)
+    if callback:
+        import importlib
+        getattr(
+            importlib.import_module(callback['module']),
+            callback['function']
+        )(user, created, attributes, ticket, service, request)
--- a/local_settings_example.py
+++ b/local_settings_example.py
@@ -210,6 +210,9 @@ RELATE_SIGN_IN_BY_EXAM_TICKETS_ENABLED = True
 # See saml_config.py.example for help.
 RELATE_SIGN_IN_BY_SAML2_ENABLED = False
 
+# If you enable this, you will also need to set up the CAS settings further down.
+RELATE_SIGN_IN_BY_CAS_ENABLED = False
+
 # }}}
 
 # {{{ editable institutional id before verification?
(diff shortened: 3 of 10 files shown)


relate/templates/sign-in-choice.html
```diff
@@ -5,6 +5,15 @@
   <h1>{% blocktrans %}Sign in to {{ relate_site_name }}{% endblocktrans %}</h1>
 
   <ul class="list-sign-in-methods">
+    {% if relate_sign_in_by_cas_enabled %}
+      <li>
+        <a
+          class="btn btn-primary"
+          href="{% url 'cas_ng_login' %}{{next_uri}}"
+          role="button"><i class="fa fa-institution"></i>
+          {% trans "Sign in using your institution's login" %} &raquo;</a>
```
inducer: In the event that both SAML and CAS are enabled, it might be useful to add a "(CAS)" here and a "(SAML)" to the SAML button.

cas-config/models.py
```diff
@@ -0,0 +1,22 @@
+from django.conf import settings
```
inducer: * It's odd to me that the name of this module is not a valid Python module name. (since there's a dash in the file name)
* It's also odd to me that this is in `models.py`, which by Django app expectations would contain database models. Could this be named `receivers.py`?
* If this is an optional part, could this live under `contrib` to avoid clutter?

local_settings_example.py
```diff
@@ -582,4 +585,21 @@
...
+    # For these settings, as well as the full list of django-cas-ng settings refer to documentation
+    # at https://github.com/mingchen/django-cas-ng
+    CAS_SERVER_URL = "https://your-cas-login-domain/cas/"
+    CAS_CREATE_USER = True
+    CAS_USERNAME_ATTRIBUTE = 'username'
+
+    # This is an extra addition. To allow one to easily use a more complex attribute mapper
+    # than a simple map in CAS_RENAME_ATTRIBUTES,
+    # we have the cas-config app, which will call the function you define here
+    # with the same parameters as django-cas-ng signal cas_user_authenticated.
```
inducer: Improve phrasing. Imagine you have no idea what this does and reading this text. :)

cas-config/.gitignore
```diff
@@ -0,0 +1 @@
+callback.py
```
inducer: Can you explain this?

inducer:
Btw, there were also a few complaints from the linter (Flake8): https://travis-ci.org/inducer/relate/jobs/520386819#L527

Could you look into addressing those, too?

cas-config/.gitignore
Mihara: Sorry for not answering earlier, we had a *long* string of holidays. I'll get back to you with the rest of it in a few hours, but I think I'll explain this one right now:

The idea is that should the user need to introduce custom code that fires on every CAS login, they would be able to do this without forking at all by dropping in a file -- presumably, here, and naming it `callback.py`. This is driven by experience: Many of our changes were entirely unsuitable for inclusion into upstream, *(chiefly because our use case was not suitable for the platform, but that's a problem not of my making)* but it was impossible to implement them without forking, which results in a maintenance nightmare.

I.e. this is where the CAS attributes delivered in an ad-hoc arbitrary format would turn into course subscriptions. I guess I'll document that...

cas-config/models.py
Mihara: > It's odd to me that the name of this module is not a valid Python module name. (since there's a dash in the file name)

I was foolishly imitating `saml-config`. Which isn't a module, come to think of it. Renamed to `cas_config` unless you can offer a better name.

> It's also odd to me that this is in models.py, which by Django app expectations would contain database models. Could this be named receivers.py?

Sorry, force of habit (we still have yet to migrate to Django 2.0). In 2.0, this can be attached by apps.py and named whatever...

> If this is an optional part, could this live under contrib to avoid clutter?

This isn't really an optional part, because it's required for the below-described `callback.py` to work if it is present. An example callback.py could be in contrib, though, which is where I'm going to put it...

relate/templates/sign-in-choice.html
Mihara: Outside the translation or inside it, though? Wouldn't want to break the translations that are already there...

Mihara:
There, took care of all the linting issues.

Mihara:
Well, since I no longer work there, there's no point keeping this around.

PR #652, opened 2019-04-22
FloatMatcher: correctly handle nan and inf

inducer:


inducer:
@davis68 Could you review this?

davis68:
I have pulled this and verified that it yields the correct behavior from the page sandbox.

PR #701, opened 2020-05-20
Added JupterNotebookUploadQuestion.

### LINKED ISSUE #690: New type of question: JupyterNotebookQuestion
Or `JupyterNotebookUploadQuestion`?
I think it's quite easy to achieve that by subclassing `UploadQuestion`. Features:
1. Allow only uploading file with extension ".ipynb"
1. Check validaty of the file
1. Enable preview in upload ui and grading ui
1. Prevent dangerous (JavaScript) code in preview

Any suggestions?

dzhuang:
Fix #690

### DIFF
--- a/course/page/upload.py
+++ b/course/page/upload.py
@@ -2,7 +2,10 @@
 
 from __future__ import division
 
-__copyright__ = "Copyright (C) 2014 Andreas Kloeckner"
+__copyright__ = """
+Copyright (C) 2014 Andreas Kloeckner
+Copyright (c) 2020 Dong Zhuang
+"""
 
 __license__ = """
 Permission is hereby granted, free of charge, to any person obtaining a copy
@@ -38,6 +41,13 @@ from relate.utils import StyledForm, string_concat
 
 from crispy_forms.layout import Layout, Field
 
+# {{{ mypy
+
+if False:
+    from typing import Optional, Text  # noqa
+
+# }}}
+
 
 # {{{ upload question
 
@@ -46,11 +56,16 @@ class FileUploadForm(StyledForm):
     uploaded_file = forms.FileField(required=True,
             label=ugettext_lazy('Uploaded file'))
 
-    def __init__(self, maximum_megabytes, mime_types, *args, **kwargs):
+    def __init__(self, maximum_megabytes, mime_types,
+                 clean_uploaded_file_callback=None,
+                 *args, **kwargs):
         super(FileUploadForm, self).__init__(*args, **kwargs)
 
         self.max_file_size = maximum_megabytes * 1024**2
         self.mime_types = mime_types
+        self.clean_uploaded_file_callback = clean_uploaded_file_callback
+        if self.clean_uploaded_file_callback is not None:
+            assert callable(self.clean_uploaded_file_callback)
 
         # 'accept=' doesn't work right for at least application/octet-stream.
         # We'll start with a whitelist.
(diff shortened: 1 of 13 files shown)


course/page/upload.py
```diff
@@ -46,11 +56,16 @@ class FileUploadForm(StyledForm):
...
+    def __init__(self, maximum_megabytes, mime_types,
+                 clean_uploaded_file_callback=None,
+                 *args, **kwargs):
         super(FileUploadForm, self).__init__(*args, **kwargs)
 
         self.max_file_size = maximum_megabytes * 1024**2
         self.mime_types = mime_types
+        self.clean_uploaded_file_callback = clean_uploaded_file_callback
+        if self.clean_uploaded_file_callback is not None:
+            assert callable(self.clean_uploaded_file_callback)
```
inducer, line 68: This should be done by subclassing the form.

course/utils.py
```diff
@@ -1241,94 +1244,114 @@ def __call__(self, *args, **kwargs):
...
+
+    from nbconvert import HTMLExporter
+    html_exporter = HTMLExporter(
+        config=config,
+        template_file=template_file
+    )
+
+    (body, resources) = html_exporter.from_notebook_node(notebook)
+
+    return "<div class='relate-notebook-container'>%s</div>" % body
```
inducer, line 1316: This is hard to review, as the code was moved. What was changed in addition to the move?

course/utils.py
```diff
@@ -1241,94 +1244,114 @@ def __call__(self, *args, **kwargs):
...
+    pages.
+    :param ipynb_source: the :class:`text` read from a ipython notebook.
+    :param indices: a :class:`list` instance, 0-based indices of notebook cells
+    which are expected to be rendered.
+    :param clear_output: a :class:`bool` instance, indicating whether existing
+    execution output of code cells should be removed.
+    :param clear_markdown: a :class:`bool` instance, indicating whether markdown
+    cells will be ignored..
+    :param config: a :class:`traitlets.config.loader.Config` instance.
+    :param config_callback: a function which further handles `config` .
```
inducer, line 1272: Be more precise here.

course/utils.py
inducer: I guess `config_callback` was added. Anything else?

tests/base_test_mixins.py
```diff
@@ -70,7 +70,7 @@
             "hidden": False,
             "listed": True,
             "accepts_enrollment": True,
-            "git_source": "git://github.com/inducer/relate-sample",
+            "git_source": "git://github.com/dzhuang/relate-sample",
```
inducer, line 73: This should be changed back before merge.

course/page/upload.py
```diff
@@ -76,15 +91,151 @@ def clean_uploaded_file(self):
...
+        if answer_data is None:
+            return None
+
+        ext = self.get_download_file_extension()
+
+        from base64 import b64decode
+        return ext, b64decode(answer_data["base64_data"])
+
+
+class FileUploadQuestion(FileUploadQuestionBase):
```
inducer, line 238: This is hard to review, as much of this code was moved. Could you outline the changes?

tests/base_test_mixins.py
dzhuang: Sure.

course/page/upload.py
dzhuang: I had that idea in the first place. Howeve, by using this callback, we do not need to do the subclassing each time we add a new type of question with different `mime_type` (I think the main difference is the clean method of the `uploaded_file`).

dzhuang: I think you are right on this.

dzhuang:
> Here's a pretty important question that occurred to me: Can you find/add a link in source to nbconvert's policy on whether it's designed to be safe with untrusted input?

I'll looking into that. At least, I found the notebook have the ability to execute Javascript scripts. I don't know whether that's still the case when its converted to html. That might be an issue before we can merge this PR.

dzhuang:
@inducer 
https://jupyter-notebook.readthedocs.io/en/stable/security.html#security-in-notebook-documents

dzhuang:
There's a [sanitize preprocessor](https://github.com/jupyter/nbconvert/blob/910c23f3225c2ced6d291424ea241d19d1077d05/nbconvert/preprocessors/sanitize.py), we need to make sure the preprocessor is included when we do the convert.

course/utils.py
dzhuang: > I guess config_callback was added. Anything else?

Yes. Because we need a different template (built-in `basic.tpl` template) as compared to what we do when we call the `render_notebook_cells` method, which uses the [`nbconvert_template.tpl`](https://github.com/inducer/relate/blob/master/course/templates/course/jinja2/nbconvert_template.tpl).

I'll check if there's other changes.

course/utils.py
dzhuang: I'll try.

course/page/upload.py
dzhuang: I'll do that.

PR #703, opened 2020-06-06
Use Py3.7 typing.TYPE_CHECKING instead of 'if False'

inducer:
@dzhuang @joshuata This makes Relate Py3.7 (and newer)-only, but it's a good idea IMO. OK with you?


inducer:
I'll proceed with this. If either of you hate it, we can revert it.

dzhuang:
I'm ok with this.  Thanks for the work!

PR #704, opened 2020-06-06
Move default git-roots from .. to a subdirectory, to avoid disasters

inducer:
(Mypy CI will need #703 to pass)


inducer:
FYI @joshuata

PR #503, opened 2018-03-18
[WIP] Pipenv

inducer:
This preserves my attempts to get pipenv to go.

Issues:

* It installs the wrong bleach
* Adapting to multiple Python versions is super hard

PR #709, opened 2020-06-08
Add poetry packaging support

joshuata:
Replaces #503 

This branch adds support for packaging using the [Poetry](https://python-poetry.org) packaging tool. It adds automatic virtualenv management as well as much stronger package requirement and resolution tools. 

The goal is for the following set of commands to start a guaranteed-compatible python environment with RELATE ready to go.

```
// clone a fresh copy of RELATE and cd into it
poetry install && poetry shell
./manage.py migrate
./manage.py runserver
```

It also automatically registers the `relate` script so that it is available on the virtualenv's path.

This currently works on my machine, but it will need more validation before it is ready to merge. 

One important feature is the ability to define dev-dependencies so that testing and coverage dependencies can be defined in the same place as all others, and allow for easy testing on local machines.


bin/relate.py
```diff
@@ -194,7 +194,7 @@ def test_code_yml(yml_file, repo_root):
             return
 
         for page in pages:
```
inducer, line 196: Not liking the rename much. This gets used as
```
relate validate .
```
or
```
relate test-code some/where.yml
```
Changing those to `relate.py` feels like it will hurt my muscle memory and add clutter for little gain.

That said, I am all for making this visible to the linter.

pyproject.toml
```diff
@@ -0,0 +1,100 @@
...
+colorama = "*"
+
+[tool.poetry.dev-dependencies]
+codecov = "^2.1.4"
+factory_boy = "^2.12.0"
+flake8 = "^3.8.3"
+pep8-naming = "^0.10.0"
+
+[tool.poetry.scripts]
+relate = "bin.relate:main"
```
inducer, line 100: > `bin/__init__.py`

`bin` isn't really intended as a module. Why is that needed?

inducer, line 100: Previously, `requirements.txt` would install enough dependencies to run the web app, whereas `setup.py` would install just enough dependencies to run the command line tool.

- Do you think that's a distinction worth keeping?
- If so, is there a way to do that with poetry?

pyproject.toml
```diff
@@ -0,0 +1,100 @@
...
+# For comfortable code entry
+django-codemirror-widget = ">=0.5"
+# For code isolation in code questions
+docker-py = "^1.10.6"
+# For code highlighting, required via the CodeHilite extension to Python-Markdown
+pygments = "^2.6.1"
+# For grade export
+unicodecsv = "^0.14.1"
+# To support network matching for facility recognition
+ipaddress = "^1.0.23"
```
inducer, line 62: My mental model was that, in addition to these version ranges, there would be something like a "lock file" (like with npm, also in pipenv), that specifies exact version numbers of all dependencies that were tested together (for more predictable deploys). Does poetry allow such a thing?

local_settings_example.py
```diff
@@ -469,6 +471,25 @@
     SAML_DJANGO_USER_MAIN_ATTRIBUTE = 'username'
     SAML_DJANGO_USER_MAIN_ATTRIBUTE_LOOKUP = '__iexact'
 
+    # The list of IdPs we talk to are defined
```
inducer, line 474:
```suggestion
```

pyproject.toml
joshuata: There is, I just haven't committed it yet since there is a lot of churn. I will as soon as its not changing drastically and things are ready to merge

bin/relate.py
joshuata: Once you run `poetry install` it install the script with the name `relate`. So it doesn't change the end-user ergonomics at all, it just moves it closer to the code it depends on and makes it visible to linting/testing. 

E.g. this works as expected on my machine
```
poetry install
poetry shell
cd relate-sample
relate validate .
```
or
```
poetry install
cd relate-sample
poetry run relate validate .
```

You can even clone the repo right now and try it out. That part of it works perfectly on my machine.

pyproject.toml
joshuata: It pulls in the script in as a native python function rather than as an opaque binary the way setuptools did it. Mostly its just the way Poetry works.

joshuata: I'm not sure that it is a useful distinction, unless you need to install the cli without the app. I can see some desire for that, but right now there is not a way to have two separate sets of dependencies like that. 

One option would be to spin out the relate course management tools (parser, validator, etc...) as a separate python package that would be consumed by the relate webapp. However that might add a bunch of overhead since there is no concept of child packages/workspaces unlike other languages like [rust](https://doc.rust-lang.org/book/ch14-03-cargo-workspaces.html)

local_settings_example.py
joshuata: Ooops. Good catch. I moved the IDP definition outside since I couldn't get flake8 to stop complaining about it.

local_settings_example.py
```diff
@@ -469,6 +471,25 @@
...
+    SAML_IDP = {
```
inducer, line 475: Since this is not directly taken in by djangosaml2, I'd say let's make this a lower case variable.

joshuata:
@inducer Any idea why there is a sudden negative change in code coverage? I've tried to double check that exactly equivalent calls are being made, but it is still showing very strange changes in the results.

run-tests-for-ci.sh
```diff
@@ -9,61 +9,66 @@ echo "-----------------------------------------------"
...
-else
-  echo "**> creating virtualenv"
-  ${VIRTUALENV} .env
-fi
+# if [ -d ".env" ]; then
+#   echo "**> virtualenv exists"
+# else
+#   echo "**> creating virtualenv"
+#   ${VIRTUALENV} .env
+# fi
```
inducer, line 34: Don't comment this out. Just kill it. We can always get it back from git.

pyproject.toml
```diff
@@ -0,0 +1,101 @@
...
+    { include = "course" },
+    { include = "accounts" },
+    { include = "relate" },
+    # { include = "bin" },
+]
+
+[tool.poetry.scripts]
+relate = "relate.bin.relate:main"
+
+[tool.poetry.dependencies]
```
inducer, line 24: This is makes `requirements.txt` redundant IMO. We shouldn't (try to) maintain both.

appveyor.yml
```diff
@@ -57,19 +57,22 @@ install:
   # Update pip/setuptools/wheel to avoid an issue like
```
inducer, line 57: There's also `.gitlab-ci.yml`, which is being used [here](https://gitlab.tiker.net/inducer/relate). Just invited you to that project. Could you also hack that into shape to use poetry?

.github/workflows/ci.yml
```diff
@@ -13,43 +13,86 @@ jobs:
...
+              uses: dschep/install-poetry-action@v1.3
+              with:
+                  create_virtualenvs: true
+            - uses: actions/cache@v1
+              id: cache
+              with:
+                  path: ~/.cache/pypoetry/virtualenvs
+                  key: pypoetry-${{ hashFiles('**/poetry.lock') }}-${{ matrix.python-version }}
+                  restore-keys: |
+                      pypoetry-${{ hashFiles('**/poetry.lock') }}-${{ matrix.python-version }}
```
inducer, line 34: Oh, cool! TIL. :)

.github/workflows/ci.yml
```diff
@@ -13,43 +13,86 @@ jobs:
...
```
inducer, line 62: A bit redundant with the above. Is there a way to reuse pieces?

pyproject.toml
```diff
@@ -0,0 +1,101 @@
+[build-system]
```
inducer, line 1: We should also update the install docs, in `doc/misc.rst`. Maybe also add a recommendation to *not* run `poetry` in an existing virtualenv. I may or may not have just been bitten by that. *cough*

inducer:
Is this ready look at? If so, please click the "request review" button to let me know. Thx!

joshuata:
I grabbed the bugfix for djangosaml2 using git, but they should release it on PyPI soon, so we can move to that as soon as it happens

inducer:
Gitlab CI is unhappy: https://gitlab.tiker.net/inducer/relate/-/merge_requests/84 (You should be able to push to that branch).

joshuata:
Fixed!

inducer:
LGTM. Thanks for working on this!

dzhuang:
Thanks!

PR #723, opened 2020-07-26
Use serializier from djangorestframework for api.

### LINKED ISSUE #719: Migrate API to django REST

dzhuang:
Fix #719 .


course/serializers.py
```diff
@@ -0,0 +1,135 @@
...
+    grader = serializers.CharField(
+        source="grader.username", read_only=True)
+
+    percentage = serializers.SerializerMethodField()
+
+    def get_percentage(self, obj):
+        if obj.correctness is not None:
+            return 100 * obj.correctness
+        else:
+            return None
```
joshuata, line 135: Do we want to make this a read-only property on the `FlowPageVisitGrade` class instead of the serializer?

dzhuang: Thanks for your review. What's your opinion? @inducer

PR #734, opened 2020-09-22
Fix empty dropdown in ModelSelect2Widget widget.

dzhuang:
After upgraded to 7.04 +, the dropdown ModelSelect2Widget needs input >=2 characters for search by default, otherwise there won't be any result. The fix was suggested by [this issue](https://github.com/applegrew/django-select2/issues/568).

### DIFF
--- a/course/auth.py
+++ b/course/auth.py
@@ -197,7 +197,10 @@ class ImpersonateForm(StyledForm):
                 queryset=qset,
                 required=True,
                 help_text=_("Select user to impersonate."),
-                widget=UserSearchWidget(queryset=qset),
+                widget=UserSearchWidget(
+                    queryset=qset,
+                    attrs={"data-minimum-input-length": 0},
+                ),
                 label=_("User"))
 
         self.fields["add_impersonation_header"] = forms.BooleanField(


inducer:
- I thought this behavior made sense, but I'm on a latin alphabet. The Illinois instance has 15,000 users, so narrowing down the query a bit seems reasonable to me. Happy to hear dissenting opinions though.
- Is there a consistency issue vs. all the other spots where select2 is used?

dzhuang:
> * I thought this behavior made sense, but I'm on a latin alphabet. The Illinois instance has 15,000 users, so narrowing down the query a bit seems reasonable to me. Happy to hear dissenting opinions though.

For Chinese, there're large amount of users with both first name and last name only one character. That means if I don't know his/her username or email or inst_id, I won't be able to find that user. :-(

> * Is there a consistency issue vs. all the other spots where select2 is used?

This behavior only affected ModelSelect2Widget, and I can confirm the impersonate form is the only use case here.

inducer:
OIC. That makes sense. Thanks for suggesting it.

dzhuang:
Thank you!

Issue #735, opened 2020-10-09
Preapproval failed to set non-default roles for user who update profile after requesting for enrollment


dzhuang:
To reproduce:

1. Create a Preapproval with inst_id "12345" which will be preapproved as role `ta`
1. An user Tom registered without inst_id, and request to enroll
1. Tom updated his inst_id as "12345" (verified)
1. Tom will be preapproved as a `student`, instead of `ta`

dzhuang:
Closed via #736

PR #736, opened 2020-10-09
Fix roles for users which update profile after enroll requests

dzhuang:
Fixing #735

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,17 +22,21 @@ jobs:
         uses: dschep/install-poetry-action@v1.3
         with:
           create_virtualenvs: true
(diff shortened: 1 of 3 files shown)


dzhuang:
Github check fails, the traceback are:

```
Python executable: The virtual environment found in /home/runner/.cache/pypoetry/virtualenvs/relate-courseware-ni3j8cUZ-py3.8 seems to be broken.
```

inducer (approved):

inducer:
The caching seems unhappy. I just disabled it.

dzhuang:
> The caching seems unhappy. I just disabled it.

Thanks for the prompt response!

PR #751, opened 2020-10-28
Added participation tags for ParticipationPreapproval.

dzhuang:
1. Added tags to ParticipationPreapproval
2. Allow set tags when preapprove.
3. Tests


dzhuang:
@inducer I think this PR is ready, and it should be merged before #745 .  (Because it requires a database migration, I think it's better that you make the decision.)

PR #758, opened 2020-12-29
Correct flow submissions

sahilkamboj3:
- Fixed bug to show 100% when student gets all answers correct
- Also, modified test_points_percent_full() and test_unreachable_points_percent_full() tests


local_settings_example.py
```diff
@@ -36,7 +36,16 @@
...
+# DATABASES = {
+#     "default": {
+#         "ENGINE": "django.db.backends.postgresql",
+#         "NAME": "relate",
+#         "USER": "sahilkamboj",
+#         "PASSWORD": 'lovebread3',
+#         "HOST": '127.0.0.1',
+#         "PORT": '5432',
+#     }
+# }
```
inducer, line 48: ?

course/flow.py
```diff
@@ -705,7 +705,7 @@ def __init__(
 
     # Rounding to larger than 100% will break the percent bars on the
     # flow results page.
```
inducer, line 707: This comment can probably be removed.

inducer, line 707: Did you take any measures to prevent the percent bars being fed numbers larger than 100 (e.g. due to rounding)?

course/migrations/0115_auto_20201226_1748.py
```diff
@@ -0,0 +1,85 @@
+# Generated by Django 3.0.11 on 2020-12-26 23:48
+
+import course.models
+import django.core.validators
+from django.db import migrations, models
+import jsonfield.fields
+
+
+class Migration(migrations.Migration):
```
inducer, line 9: Where did this migration come from?

inducer:
Thanks for working on this! I would argue that your changes don't really address the issue, as the rounded values could very well still add up to a value greater than 100.

Observe:

- 0.6
- 0.6
- 0.6

Each round to 1. Suppose 2 (instead of 100) is the upper limit. Individually, the sum is 1.8. Rounded, they sum to 3.

course/templates/course/flow-completion-grade.html
```diff
@@ -42,12 +42,18 @@ <h1>{% trans "Results" %}: {{flow_desc.title}}</h1>
                   {% if grade_info.total_points_percent < 100.001 %}
                       {# otherwise we'll have trouble drawing the bar #}
                       <div class="progress">
+                          {% with points_percent=grade_info.points_percent|mul:0.9999 %}
```
inducer, line 45: Add a template `{# ... #}` comment on why that's here.

course/templates/course/flow-completion-grade.html
```diff
@@ -64,14 +70,22 @@ <h1>{% trans "Results" %}: {{flow_desc.title}}</h1>
                       {% endif %}
                   </p>
                   <div class="progress">
+                      {% with fully_correct_percent=grade_info.fully_correct_percent|mul:0.9999 %}
```
inducer, line 73: Add a template `{# ... #}` comment on why that's here.

course/templates/course/flow-completion-grade.html
```diff
@@ -89,14 +103,22 @@ <h1>{% trans "Results" %}: {{flow_desc.title}}</h1>
                       {% endif %}
                   </p>
                   <div class="progress">
+                      {% with optional_fully_correct_percent=grade_info.optional_fully_correct_percent|mul:0.9999 %}
```
inducer, line 106: Add a template `{# ... #}` comment on why that's here.

inducer (commented):
Thanks! (And sorry again about the epic delay... life got busy in a hurry.) This looks great. My only request is to add comments explaining why we're doing the dodgy-looking multiplication, so that folks who read this code in the future will have an easier time understanding. After that, it's ready to land.

PR #765, opened 2021-01-31
Identical code option added

sahilkamboj3:


course/page/code.py
```diff
@@ -593,6 +593,8 @@ def __init__(self, vctx, location, page_desc, language_mode):
                     "While you're at it, consider adding "
                     "access_rules/add_permssions/see_correctness."))
 
+        self.identical_okay = getattr(page_desc, "identical_okay", False)
```
inducer, line 596: Needs to be added to the documentation.

inducer, line 596: Naming? `exact_correct_code_ok`?

inducer (commented):
Thanks! Two thoughts below.

Issue #767, opened 2021-02-03
Windows path handling is broken


inducer:
As discussed [here](https://github.com/inducer/relate/pull/556#issuecomment-772746240).

cc @dzhuang

inducer:
In particular, that means we shouldn't use `os.path` functions for path processing.

inducer:
The way I understand path handling in Relate on Windows right now is that one would use `\\` as a path separator (e.g. in Jinja includes), whereas one would use `/` on Linux. This means that you can't use a Linux-based Relate repo on a Windows Relate and vice versa. Is my understanding accurate here? If so, IMO this is a problem that should be easy to avoid, perhaps best by standardizing on `/` as a path separator (and hence not using `os.path` to process paths, which uses system-dependent separators).

dzhuang:
Thanks for your explanation, I think your suggestion is reasonable because my patch might result in failure. However, In my case, all my repos can work on both on Windows (for debugging) and Linux (for production). Can you provide the traceback which lead to this issue? I want to create a testcase which actually fix that, thanks.

inducer:
In your repos, what do you use for path separators, e.g. when using Jinja:
```
{% from "path1/path2/yaml-macros.jinja" import hw_header %}
```
?
(You may not have any of those that have directory components in them. In that case, please test them, because I strongly suspect they're broken.)

dzhuang:
Got it. I'll have a try.

dzhuang:
> In your repos, what do you use for path separators, e.g. when using Jinja:

I used `/` in my repos. FYI, I only `\` when configuring `GIT_ROOT` in local_settings, e.g., `GIT_ROOT = r"D:\document\python\course\course-git"` on Windows.

For this kind of snippet:
```
{% from "path1/path2/yaml-macros.jinja" import hw_header %}
```

 I can do macro import from 2-level subfolder without issue both on Windows and Linux. Am I missing anything?

dzhuang:
> In your repos, what do you use for path separators, e.g. when using Jinja:

I used `/` in my repos. FYI, I only `\` when configuring `GIT_ROOT` in local_settings, e.g., `GIT_ROOT = r"D:\document\python\course\course-git"` on Windows.

For this kind of snippet:
```
{% from "path1/path2/yaml-macros.jinja" import hw_header %}
```

 I can do macro import from 2-level subfolder without issue both on Windows and Linux. Am I missing anything?

inducer:
```
[...]

PR #783, opened 2021-03-25
Added tests for #556.

dzhuang:
Testing #556 as mentioned in #767. The tests works without problems. The merge of this PR requires https://github.com/dzhuang/relate-sample/commit/f43b734c85de57fbab262ba5a19d0fe9d8733a4d merged into sample repo. 

P.S., We should metion in docs that path split should always use `/`.

@inducer

### DIFF
--- a/tests/base_test_mixins.py
+++ b/tests/base_test_mixins.py
@@ -69,7 +69,7 @@ SINGLE_COURSE_SETUP_LIST = [
             "hidden": False,
             "listed": True,
             "accepts_enrollment": True,
-            "git_source": "git://github.com/inducer/relate-sample",
+            "git_source": "git://github.com/dzhuang/relate-sample",
             "course_file": "course.yml",
             "events_file": "events.yml",
             "enrollment_approval_required": False,
--- a/tests/test_content.py
+++ b/tests/test_content.py
@@ -853,7 +853,8 @@ class YamlBlockEscapingGitTemplateLoaderTest(SingleCourseTestMixin, TestCase):
             with self.pctx.repo as repo:
                 loader = content.YamlBlockEscapingGitTemplateLoader(
                     repo, self.course.active_git_commit_sha.encode())
-                result = loader.get_source(environment, "content-macros.jinja")
+                result = loader.get_source(environment,
+                                           "macros/content-macros.jinja")
                 source, _, _ = result
                 self.assertIsNotNone(source)
                 self.assertTrue(source.startswith(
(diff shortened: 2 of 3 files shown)


tests/base_test_mixins.py
```diff
@@ -69,7 +69,7 @@
             "hidden": False,
             "listed": True,
             "accepts_enrollment": True,
-            "git_source": "git://github.com/inducer/relate-sample",
+            "git_source": "git://github.com/dzhuang/relate-sample",
```
dzhuang, line 72: This need to be reverted when merging.

inducer:
Thanks for doing this! I really don't understand why this works though. I tried whether dulwich does path parsing:
```
from dulwich.repo import Repo
r = Repo(".")
c = r[b"HEAD"]
t = r[c.tree]
f = t[b"pytools/__init__.py"]
print(f)
```
(run in a checkout of https://github.com/inducer/pytools/, because why not) It does not.

So if this code:

https://github.com/inducer/relate/blob/f4e8dee04dd8ac67d724412fe86957404ccb98cc/course/content.py#L651-L653

splits the path by `os.sep` (which i assume is `\` in Windows, right?), how does it feed the right file names to dulwich? (I don't have a Windows system to check, otherwise I would)

dzhuang:
OIC. I finally remembered the reason for proposing #556.

The `.attribute` file won't be parsed correctly in Windows. e.g., the file `images/.attributes.yml` will be `images\.attributes.yml` in Windows.

Now I can see this line (`os.path.join`)lead to that result:
https://github.com/inducer/relate/blob/f4e8dee04dd8ac67d724412fe86957404ccb98cc/course/content.py#L803

I didn't realized that when I try to fix this issue, and in #556 the workround went to a wrong direction. Although it actually fixed it, I must admit it's bad and sorry for  causing your confusion.

So the better workround is to use `"/".join(os.path.dirname(path), ATTRIBUTES_FILENAME)`, am I right?

dzhuang:
FYI, `os.path.normpath`
```python
>>> os.path.normpath("repo/root/sub")
'repo\\root\\sub'
>>> os.path.normpath(r"repo/root/sub\.attribute") 
'repo\\root\\sub\\.attribute'
```

`normpath` will replace all `\` and `/` to `os.sep`, and the we can use `os.sep` to split the path to elements.

inducer:
I propose that we standardize on `/` as a path separator in course content. I'm not a fan of allowing both (as we would if we used `normpath`). I agree that using `os.path.join` to construct repo paths (in the spot you found, but there are a few more) is incorrect, we should fix those instances.

How do you feel about this direction?

dzhuang:
I agree. But my first attemp failed (see #784). Maybe there're more `os.path.join` need to be changed as you mentioned.

PR #784, opened 2021-03-29
Actually fix path join in Windows.

### LINKED ISSUE #767: Windows path handling is broken

dzhuang:
Reverting changes in #556 and fix #767 

- [ ] Adding tests.

### DIFF
--- a/course/content.py
+++ b/course/content.py
@@ -648,9 +648,7 @@ def look_up_git_object(repo: "dulwich.Repo",
         raise ObjectDoesNotExist(_("symlink nesting depth exceeded "
             "while locating '%s'") % full_name)
 
-    # https://github.com/inducer/relate/pull/556
-    # FIXME: https://github.com/inducer/relate/issues/767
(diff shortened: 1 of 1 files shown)


course/content.py
```diff
@@ -648,9 +648,7 @@ def look_up_git_object(repo: "dulwich.Repo",
         raise ObjectDoesNotExist(_("symlink nesting depth exceeded "
             "while locating '%s'") % full_name)
 
-    # https://github.com/inducer/relate/pull/556
-    # FIXME: https://github.com/inducer/relate/issues/767
-    name_parts = os.path.normpath(full_name).split(os.sep)
+    name_parts = list(os.path.split(os.path.normpath(full_name)))
```
dzhuang, line 651: The test fails because `os.path.split` result in a 2-tuple, not split the path into path elements.
However, change to 
```
name_parts = os.path.normpath(full_name).split("/")
```
also fails in Windows.

@inducer

PR #791, opened 2021-04-16
Add course flag to decide whether arbitrary HTML is allowed

inducer:
- ~~Add "trusted instructor" permission that defaults to construction of couses with markup allowed~~
- [x] Migration: Set `trusted_for_markup` on existing courses.


inducer:
FYI @davis68 When creating a new course after this is merged, arbitrary HTML in course content will require "may present arbitrary HTML"  setting (`trusted_for_markup`) for the course (in the admin).

PR #803, opened 2021-04-21
Add Github CodeQL

inducer:


inducer:
Hmm... not sure what it does :shrug: but it can't hurt.

PR #909, opened 2022-03-20
Fix locale loading and late events rendering in calendar.

### LINKED ISSUE #907: Issue with fullCalendar
1. Locale doen't work. There needs a param for `locale` （beside `locales`) in `SetupCalendar`.

https://github.com/inducer/relate/blob/0f1bbd05b62e385981b6711e99c21f6a1efc6a6b/relate/static/js/fullcalendar.js#L8-L26

2. Events started after 23:00 pm will span 2 days when rendering. https://github.com/fullcalendar/fullcalendar/issues/6265

dzhuang:
Fix #907 
1. Added locale param in `setupCalendar`
2. Set [`defaultTimedEventDuration`](https://fullcalendar.io/docs/defaultTimedEventDuration) to 1 ms
3. Replace Chinese specific locale name when build.

### DIFF
--- a/relate/static/js/fullcalendar.js
+++ b/relate/static/js/fullcalendar.js
@@ -5,7 +5,7 @@ import listPlugin from '@fullcalendar/list';
 import allLocales from '@fullcalendar/core/locales-all';
 
 /* eslint-disable-next-line import/prefer-default-export */
-export function setupCalendar(domEl, events, initialDate) {
+export function setupCalendar(domEl, events, initialDate, locale) {
   const calendar = new Calendar(domEl, {
     plugins: [
       dayGridPlugin,
@@ -13,6 +13,7 @@ export function setupCalendar(domEl, events, initialDate) {
       listPlugin,
     ],
     locales: allLocales,
+    locale,
     initialView: 'dayGridMonth',
     headerToolbar: {
       left: 'prev,next today',
@@ -21,6 +22,7 @@ export function setupCalendar(domEl, events, initialDate) {
     },
     initialDate,
     events,
+    defaultTimedEventDuration: '00:00:00.001',
   });
   calendar.render();
 }
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -5,6 +5,7 @@ import commonjs from '@rollup/plugin-commonjs';
 import { terser } from 'rollup-plugin-terser';
 import styles from 'rollup-plugin-styles';
 import gzipPlugin from 'rollup-plugin-gzip';
+import replace from '@rollup/plugin-replace';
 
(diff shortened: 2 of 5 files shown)


rollup.config.js
```diff
@@ -22,6 +23,11 @@ const defaultPlugins = [
     customCompression: (content) => brotliPromise(Buffer.from(content)),
     fileName: '.br',
   }),
+  replace({
+      'zh-cn': 'zh-hans',
+      'zh-tw': 'zh-hant',
+      preventAssignment: true,
+  })
```
inducer, line 30: IIUC, this amounts to a global search-replace of these strings. That seems exceedingly risky. Why not just do a local replace in the code that passes `LANGUAGE_CODE`?

relate/static/js/fullcalendar.js
```diff
@@ -21,6 +22,7 @@ export function setupCalendar(domEl, events, initialDate) {
     },
     initialDate,
     events,
+    defaultTimedEventDuration: '00:00:00.001',
```
inducer, line 25: Wouldn't this render events that specify no end time as really narrow slivers? (Could you post a screenshot of an event that makes use of this?)

inducer (commented):
Thanks for working on this. A couple comments below.

rollup.config.js
dzhuang: I would agree with that, then we might need to revert https://github.com/inducer/relate/commit/05f95361dd4737ac06ae61e6c3d9284457f227ca#diff-3b26411dfcb9fd30a16787b21f7786d3e24c169058aedae6bc8a3e86efe87ecb back?

relate/static/js/fullcalendar.js
dzhuang: It works well after this patch in `dayGridMonth` view and `listWeek`, and as you expected, the sliver is narrow in the `timeGridWeek` view. 

**Before patched:**
5a8cca94fddc130dd71e34ad3197d9a

34718bfa01985ac0898b9b106299cd0

d67304f45b5ecf13fe616923f64df76

--------------------------------------------------------------------------------------------------------------------------------
**After patched:**

b525113195b2dd1ca49b6d3ec4dd068

58402f39e6b5f64cc2330cec59cf8e3

e062a5fd531835d0a416b0a8615a6c0


I had tried to add that `defaultTimedEventDuration` option only to `dayGridMonth` view, however, it failed because the option is not functioning as a (`view-specific-option`)[https://fullcalendar.io/docs/view-specific-options]. Any suggestions?

rollup.config.js
inducer: How about #913?

PR #913, opened 2022-03-22
Fix calendar locale

inducer:
@dzhuang, how's this?


dzhuang:
I'm ok with this. Thanks for the work!

PR #921, opened 2022-03-28
add docker build and helm chart

KlavsKlavsen:
First edition of relate Helm chart and Dockerfile - "works for me" mode :)
Pls review, ask and we can take it from there.
NB. This is based off of https://github.com/bitnami-labs/sealed-secrets/tree/main/helm/sealed-secrets - and I've removed a lot of irrelevant stuff and added some.. but the original license still applies I'd say. I prefer GPLv3 myself - but either will do here.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,0 +1,32 @@
+FROM ubuntu:20.04
+MAINTAINER Klavs Klavsen <klavs@enableit.dk>
+
+WORKDIR /tmp
+
+RUN apt-get update &&  \
+    apt-get install -y \
+    python3-pip \
+    python3.8-venv \
+    build-essential \ 
+    postgresql-server-dev-all
+
+RUN DEBIAN_FRONTEND=noninteractive TZ=Etc/UTC apt-get install git gettext-base curl -y
+RUN curl -fsSL https://deb.nodesource.com/setup_16.x | bash -
+RUN apt-get update; DEBIAN_FRONTEND=noninteractive apt-get install -y nodejs 
+
+WORKDIR /var/www
+
+RUN python3 -m venv relate-venv
+RUN pip install poetry
+RUN git clone https://github.com/inducer/relate.git
+ENV VENV_PATH=/var/www/relate-venv
+WORKDIR /var/www/relate
+RUN git pull
+RUN poetry install
+RUN npm install ; npm run build
+RUN poetry run pip install psycopg2
+COPY local_settings_template.py /var/www/relate/local_settings_template.py 
+COPY run-relate.sh /run-relate.sh
+RUN chmod 755 /run-relate.sh
+
+CMD ['/run-relate.sh']
--- a/docker/local_settings_template.py
+++ b/docker/local_settings_template.py
@@ -1,0 +1,625 @@
+# See https://docs.djangoproject.com/en/dev/howto/deployment/checklist/
+
+import os.path as path
+
+_BASEDIR = path.dirname(path.abspath(__file__))
+
+# {{{ database and site
+
+# TODO: Need to find out WHY this needs to be set in production - and ensure its set from a secret - so it stays the same
+SECRET_KEY = "<CHANGE ME TO SOME RANDOM STRING ONCE IN PRODUCTION>"
+
+ALLOWED_HOSTS = [ $EXTRAHOSTS ]
+
+CSRF_TRUSTED_ORIGINS = [ $ORIGINS ]
+
+# Configure the following as url as above.
+RELATE_BASE_URL = "https://relate.obmondo.com"
+
+from django.utils.translation import gettext_noop  # noqa
+
+# Uncomment this to configure the site name of your relate instance.
+# If not configured, "RELATE" will be used as default value.
+# Use gettext_noop() if you want it to be discovered as an i18n literal
+# for translation.
+#RELATE_CUTOMIZED_SITE_NAME = gettext_noop("My RELATE")
+
+# Uncomment this to use a real database. If left commented out, a local SQLite3
+# database will be used, which is not recommended for production use.
+#
+DATABASES = {
+     "default": {
+         "ENGINE": "django.db.backends.postgresql",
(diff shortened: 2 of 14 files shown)


KlavsKlavsen:
Well - the license is right there in the repo: https://github.com/bitnami-labs/sealed-secrets/blob/main/LICENSE
That sounds like a very odd and illogical attitude towards a license - which is merely one giving us more rights with it - than we would have, was it not given (as it would then be copyright rights only).
Whomever decides to grant code under whatever license they deem fit - is IMHO just very nice, as long as it fits with my goals for using the work in question ofcourse - otherwise its just unusable. (like Oracle/Sun's CDDL license of ZFS for linux.. that means it can't be merged into Linux. sad).

inducer:
> Well - the license is right there in the repo:

Ah. Didn't realize I was looking at a subfolder. Thanks!

> That sounds like a very odd and illogical attitude towards a license

I'm sorry you feel that way! I'll quickly try to explain why I try to keep my licensing liberal:

- The more users something has, the better.
- The virality of the GPL makes GPL'd software high-risk for corporate users. (That's deliberate, as far as I can tell.)
- Personally, I'd prefer (voluntary) contributions from corporate folks based on gratitude over no contributions from no users because of onerous license restrictions.
- And if somebody gets rich off of what I've done, but doesn't want to share... that's fine, too. :shrug:

KlavsKlavsen:
I have never been in a company that did not highly appreciate contributing to GPL projects. 
Actually MANY projects PREFER GPL - as they are then CERTAIN the contributions they share - cannot be turned around and sold as a closed-source product - where they don't get others contributions back.
And I've worked in large Danish orgs - and in my mind - that makes perfect sense - as it evens the playing field for all contributors, so the collaboration can be about the technical needs - and no one can turn around and "build their own improved edition they sell as closed source".

KlavsKlavsen:
I can understand the attitude as an individual - and it highly depends on wether or not you're building a "library" or not.. (thats why there's the LGPL from that family of licenses ;) - a project such as relate is a standalone product - so its not really something I would worry about "infecting" other software - and IMHO it makes only sense to run on 100% Open Source - and when we don't live of selling software - there will never be any "infection problem" for us - and we are free to focus on upstreaming all improvements we do, first (everyone here is told to do so) - and we, and the customers we help, benefit highly from others doing the same.
Frankly if all that fantastic open source software did not exist, pretty much ALL the companies I've worked for, would have been far worse off - and not be able to operate at such low cost as they do. This does mean most companies do not have enough people in operations to be able to afford to contribute back (or even to do things properly IMHO) - which is why we built our service, to share the open source work amongst all customes - who share 80-90% of the software - and thusly can share in the cost of improving it and the setup its used in.
I'm not a huge fan of the business model of selling a "closed source product" - but every company must find a way to exist - but I have no doubt that the large orgs (google, apple, oracle etc.) will crap on everyone, if it makes them more money - and a license such as the GPL helps protect us smaller companies - which are the ones that actually pay taxes in the countries they exist in - as opposed to the large ones which IMHO are very sad companies that no longer can see over "the huge sums of money they bring in to the owners".

inducer:
Thanks for working on this! Is there a way to exercise this in CI? I ask because I think it would be useful to notice in case it gets broken.

KlavsKlavsen:
I know some spin up k3s in CI - as its lightweight enough - to be used for testing - so that should be possible.

KlavsKlavsen:
We finally got Relate up and running for us again.. and now have new usecases for internal education :)

A colleague will be adding improvements we've made for this - to this PR.

I've also clarified that upstream is Apache 2.0 license - so that can be used or GPLv3 at your perogative. This is allowed as explained here: https://www.apache.org/licenses/GPL-compatibility.html

KlavsKlavsen:
I was thinking of using this action to test helm chart. Would that be acceptable to you? https://github.com/helm/chart-testing-action

inducer:
Yes, that sounds perfectly fine. I'm just worried about adding code that isn't tested, because then I have no way to even tell whether it has bitrotted.

helm/relate/readme.md
inducer, line 1: Could you add a bit in the actual documentation (under `doc/`) that points to this README? (i.e. where it would end up living in the Relate repo)

docker/local_settings_template.py
inducer, line 1: It seems ruff is a bit irritated by the syntax in this file. Could you hide it from the linter?

docker/Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+RUN  python3 -m venv relate-venv
+RUN DEBIAN_FRONTEND=noninteractive apt-get install -y python3-poetry
+
+
+RUN git clone https://github.com/inducer/relate.git
+ENV VENV_PATH=/var/www/relate-venv
+WORKDIR /var/www/relate
+RUN git pull
+RUN poetry install
+RUN npm install ; npm run build
```
inducer, line 27: See `update.sh` in the project root for a concise way to run all required steps.

docker/Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+RUN poetry run pip install psycopg2
```
inducer, line 28:
```suggestion
```

docker/Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+WORKDIR /var/www
+RUN  python3 -m venv relate-venv
+RUN DEBIAN_FRONTEND=noninteractive apt-get install -y python3-poetry
+
+
+RUN git clone https://github.com/inducer/relate.git
+ENV VENV_PATH=/var/www/relate-venv
+WORKDIR /var/www/relate
+RUN git pull
+RUN poetry install
```
inducer, line 26:
```suggestion
RUN poetry install -E postgres -E memcache
```

inducer, line 26: On that note, memcached can be helpful (though it is not required).

inducer (commented):
Thanks for working on this. Some comments below.

PR #946, opened 2022-08-05
Remove non exist static references.

dzhuang:
Since the introduction of `django-bootstrap-datepicker-plus`, those links were no longer valid.

### DIFF
--- a/course/templates/course/generic-course-form.html
+++ b/course/templates/course/generic-course-form.html
@@ -10,11 +10,6 @@
 
 {# keep this in sync with relate/templates/generic-form.html #}
 {% block head_assets_form_media %}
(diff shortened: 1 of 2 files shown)


inducer:
Thanks! These changes are already part of #892, which I am hoping to merge over the next week or so. As a result, I think I would prefer to merge them via that route. In fact, the only thing that's missing in #892 from my perspective are some rendering issues with the `InlineMultiQuestion`, specifically the various tooltip and right/wrong indicators. If you are able, it would be super helpful if you could take a look at that since you know that code better than me, and, perhaps, submit a PR targeting the `bs5` branch.

dzhuang:
Thanks. I guess the issue you mentioned is the removal of `glyphicons` in Bootstrap 5, I'll try to work on that in a few days.

inducer:
It's not so much the glyphicons, it's all the popup/over/under logic and the field layout which has changed substantially in bs5.

dzhuang:
This PR is closed in favor of #949.

PR #1023, opened 2023-09-09
Add a yaml linting subcommand

a-alveyblanc:
Replacement for running `yamllint` from the command-line, particularly in the case of CI.
- Expands yaml files (i.e. expands any Jinja found in yaml files), then performs linting
- Config file defaults to `./.yamllint`
- Requires a directory to be supplied (could change this to run in the repo root?)

LMK what you think! Seems sort of barebones, maybe I'm missing some things.

EDIT: Just realized this probably needs a test :sweat_smile:


relate/bin/relate.py
```diff
@@ -305,6 +328,12 @@ def main() -> None:
     parser_expand_yaml.add_argument("YAML_FILE")
     parser_expand_yaml.set_defaults(func=expand_yaml_ui)
 
+    parser_lint_yaml = subp.add_parser("lint_yaml")
+    parser_lint_yaml.add_argument("--repo-root", default=os.getcwd())
+    parser_lint_yaml.add_argument("--config-file", default="./.yamllint")
+    parser_lint_yaml.add_argument("DIRECTORY")
```
inducer, line 334: I'd keep this lower case. (You can set `metavar=` to have it shown upper case in `--help`.)

relate/bin/relate.py
```diff
@@ -91,8 +91,31 @@ def expand_yaml(yml_file, repo_root):
...
+    from yamllint.config import YamlLintConfig
+
+    conf = YamlLintConfig(file=args.config_file)
+
+    for root, _, filenames in os.walk(args.DIRECTORY):
+        for f in filenames:
+            filepath = os.path.join(root, f)
+            if not conf.is_file_ignored(f) and conf.is_yaml_file(f):
+                # expanded yaml is missing a newline at the end of the file
+                # which causes the linter to complain, so we add a newline :)
```
inducer, line 109: Is that newline missing for all Jinja output?

relate/bin/relate.py
inducer, line 334: Maybe allow more than one, and allow both files and directories?

relate/bin/relate.py
```diff
@@ -91,8 +91,31 @@ def expand_yaml(yml_file, repo_root):
 # }}}
 
 
-# {{{ code test
+# {{{ lint YAML
+def lint_yaml(args):
```
inducer, line 95: Could you add `yamllint` to `pyproject.toml`? (and re-lock the dependencies)

inducer (commented):
Thanks! This is already very close to what I was thinking. A test that exercises this would be welcome, but no need to overengineer it. But do make sure it runs in both Gitlab and Github.

relate/bin/relate.py
a-alveyblanc: The newline is missing for any files that are processed by `expand_yaml`. The linter will complain that the yaml file is missing a newline at the end of the file.

relate/bin/relate.py
```diff
@@ -305,6 +343,14 @@ def main() -> None:
     parser_expand_yaml.add_argument("YAML_FILE")
     parser_expand_yaml.set_defaults(func=expand_yaml_ui)
 
+    parser_lint_yaml = subp.add_parser("lint_yaml")
+    parser_lint_yaml.add_argument("--repo-root", default=os.getcwd())
+    parser_lint_yaml.add_argument("--config-file", default="./.yamllint")
+    parser_lint_yaml.add_argument("--directories", metavar="DIRECTORIES",
+                                  nargs="*")
+    parser_lint_yaml.add_argument("--files", metavar="FILES", nargs="*")
```
inducer, line 351: No need to force the user to keep them separate: `os.path.isdir`. :slightly_smiling_face:

a-alveyblanc:
Just added two simple test cases. This is GTG from my end

inducer:
That test did not do much. FTFY, see pushed commit(s).

Issue #973, opened 2022-09-29
Sample course content broken with restricted HTML


inducer:
Specifically:

- the Youtube include
- the Notebook render

PR #1036, opened 2023-10-20
feat: whitelist iframe/object tags

unRARed:
Relates to #973. Gets iframe/objects rendering again. I realize there's security concerns around iframes. Maybe this should be conditional based on an ENV var, so the user needs to opt in?

Screenshot 2023-10-20 at 8 27 57 AM

### DIFF
--- a/course/content.py
+++ b/course/content.py
@@ -1383,6 +1383,10 @@ def filter_html_attributes(tag, name, value):
                 or (name == "class" and value.startswith("btn btn-")))
     elif tag == "img":
         result = result or name == "src"
+    elif tag == "iframe":
+        result = (result or (name in ["src", "width", "height"]))
+    elif tag == "object":
+        result = (result or (name in ["data", "width", "height"]))
     elif tag == "div":
         result = result or (name == "class" and value == "well")
     elif tag == "i":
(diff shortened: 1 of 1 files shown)


inducer:
Cf. #791, i.e. there is a checkbox (only visible on the admin side) to determine whether the a course may show arbitrary HTML. This is part of a larger push towards changing the trust model so that course authorship does not require the instructor to be "trustworthy", so that this could be opened up to the wider world.

Allowing iframes would effectively revert to the state pre-#791; we might as welll allow arbitrary HTML at that point.

unRARed:
I see and it's [documented here](https://documen.tician.de/relate/content.html#in-line-html) even. Somehow I overlooked it. Maybe I overlooked this, too, but is there a convention for custom styling? I found `relate/static/css`, but I don't see that being rendered anywhere. Thanks for the awesome work you've done with this project! Love the git-based approach!

inducer:
To discuss custom styling, could you start a new discussion in the "Discussion" area and:

- explain a bit more would you like? In general, you can include CSS in your HTML and it'll work (and centralize that via Jinja)
- or you would like to change the appearance of the site wholesale?

inducer:
Thanks for the contribution, closing here since it seems your concern was addressed. LMK if that's not the case.

unRARed:
Once I get some content going for my instance I might play around with a more maintainable setup. For now, I ended up just putting a style block in base.html.

PR #1047, opened 2024-02-14
ipython notebook uploads

lukeolson:
- add ipynb mimetype
- check that json loads
- check that json has `nbformat` >= 4

### DIFF
--- a/course/page/upload.py
+++ b/course/page/upload.py
@@ -70,9 +70,21 @@ class FileUploadForm(StyledForm):
                     % {"allowedsize": filesizeformat(self.max_file_size),
                         "uploadedsize": filesizeformat(uploaded_file.size)})
 
-        if self.mime_types is not None and self.mime_types == ["application/pdf"]:
-            if uploaded_file.read()[:4] != b"%PDF":
-                raise forms.ValidationError(_("Uploaded file is not a PDF."))
+        if self.mime_types is not None:
+            if self.mime_types == ["application/pdf"]:
+                if uploaded_file.read()[:4] != b"%PDF":
+                    raise forms.ValidationError(_("Uploaded file is not a PDF."))
+            if self.mime_types == ["application/x-ipynb+json"]:
+                try:
+                    # make sure it is loadable json
+                    import json
+                    data = json.load(uploaded_file)
+
+                    # check for a notebook format of at least 4
+                    assert int(data["nbformat"]) >= 4
+                except BaseException as e:
+                    raise forms.ValidationError(_("Uploaded file is not a "
+                                                  "Jupyter notebook.")) as e
 
         return uploaded_file
 
@@ -126,6 +138,7 @@ class FileUploadQuestion(PageBaseWithTitle, PageBaseWithValue,
         * ``application/pdf`` (will check for a PDF header)
         * ``text/plain`` (no check performed)
         * ``application/octet-stream`` (no check performed)
+        * ``application/x-ipynb+json`` (will check for JSON and nbformat>=4)
 
     .. attribute:: maximum_megabytes
 
@@ -150,6 +163,7 @@ class FileUploadQuestion(PageBaseWithTitle, PageBaseWithValue,
             "application/pdf",
             "text/plain",
             "application/octet-stream",
+            "application/x-ipynb+json",
             ]
 
     def __init__(self, vctx, location, page_desc):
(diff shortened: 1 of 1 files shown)


course/page/upload.py
```diff
@@ -70,9 +70,21 @@ def clean_uploaded_file(self):
...
                         "uploadedsize": filesizeformat(uploaded_file.size)})
 
-        if self.mime_types is not None and self.mime_types == ["application/pdf"]:
-            if uploaded_file.read()[:4] != b"%PDF":
-                raise forms.ValidationError(_("Uploaded file is not a PDF."))
+        if self.mime_types is not None:
+            if self.mime_types == ["application/pdf"]:
+                if uploaded_file.read()[:4] != b"%PDF":
+                    raise forms.ValidationError(_("Uploaded file is not a PDF."))
+            if self.mime_types == ["application/x-ipynb+json"]:
```
inducer, line 77:
```suggestion
            elif self.mime_types == ["application/x-ipynb+json"]:
```

course/page/upload.py
```diff
@@ -70,9 +70,21 @@ def clean_uploaded_file(self):
...
+                try:
+                    # make sure it is loadable json
+                    import json
+                    data = json.load(uploaded_file)
+
+                    # check for a notebook format of at least 4
+                    assert int(data["nbformat"]) >= 4
+                except BaseException as e:
```
inducer:
```suggestion
                except Exception as e:
```

course/page/upload.py
```diff
@@ -70,9 +70,21 @@ def clean_uploaded_file(self):
...
+                if uploaded_file.read()[:4] != b"%PDF":
+                    raise forms.ValidationError(_("Uploaded file is not a PDF."))
+            if self.mime_types == ["application/x-ipynb+json"]:
+                try:
+                    # make sure it is loadable json
+                    import json
+                    data = json.load(uploaded_file)
+
+                    # check for a notebook format of at least 4
+                    assert int(data["nbformat"]) >= 4
```
inducer, line 84: Don't use `assert` for input checking. (might be inadvertently disabled via `-O`)

course/page/upload.py
```diff
@@ -294,7 +308,8 @@ def normalized_bytes_answer(self, page_context, page_data, answer_data):
 
         subm_data, subm_mime = self.get_content_from_answer_data(answer_data)
 
-        from mimetypes import guess_extension
+        from mimetypes import add_type, guess_extension
+        add_type('application/x-ipynb+json', '.ipynb')
```
inducer: Maybe move this to the module level? There are other uses of `guess_extension` that might also benefit.

course/page/upload.py
lukeolson: good point.  added a new raise, but take a look

course/page/upload.py
lukeolson: done.  it's at the top top

course/page/upload.py
inducer: Is that valid syntax? What does it raise?

lukeolson: it is probably better to raise a specific type of error.  added a `ValueError`

inducer:
LMK when this is ready to look at/merge. So far, I'm refraining from taking a  detailed look while it's still in draft status.

lukeolson:
> LMK when this is ready to look at/merge. So far, I'm refraining from taking a detailed look while it's still in draft status.

Go ahead now.  I've tested the functionality locally.  If a unit test would help, I'll add.

course/page/upload.py
```diff
@@ -294,7 +310,7 @@ def normalized_bytes_answer(self, page_context, page_data, answer_data):
 
         subm_data, subm_mime = self.get_content_from_answer_data(answer_data)
 
-        from mimetypes import guess_extension
+        add_type("application/x-ipynb+json", ".ipynb")
```
inducer, line 313: Could you move this into the module root?

inducer (commented):
LGTM. Just one wrinkle, then this should be good.

course/page/upload.py
lukeolson: The import is already at the module level.  Are you also wanting `add_type()` called at the module root?

inducer: Yes, so that there's not a behavior change for the other uses of `mimetypes` based on whether this method has been called.

PR #1093, opened 2024-09-10
Add PrairieTest integration

inducer:
- [x] Add (undocumented) `if_has_prairietest_exam_access` permission rule
- [x] Tell the user what URL to use, be noisy about trailing slash.
- [x] Make sure `secret` doesn't appear blank.
- [x] Make sure inline sig check test runs
- [x] Add tests of deny event checks
- [x] Add tests of allow event checks


prairietest/utils.py
```diff
@@ -0,0 +1,204 @@
...
+
+def has_access_to_exam(
+            course: Course,
+            user_uid: str,
+            exam_uuid: str,
+            now: datetime,
+            ip_address: IPv4Address | IPv6Address
+        ) -> None | AllowEvent:
+    # NOTE: This assumes that there aren't two exams with the same UUID
+    # across multiple testing centers. Otherwise, allow events could be missed.
```
inducer, line 125: This may cause a problem with the Chicago CBTF.

inducer: Now fixed.

PR #1132, opened 2024-10-17
Import dark theme from https://github.com/dennis84/codemirror-themes/…

peepeetee:
…blob/main/theme/github-dark.ts

Import dark theme from https://github.com/dennis84/codemirror-themes/blob/main/theme/github-dark.ts

The correct solution here is to go back to https://discuss.codemirror.net/t/dynamic-light-mode-dark-mode-how/4709/5 and look at the solution: https://discuss.codemirror.net/t/dynamic-light-mode-dark-mode-how/4709/2

The idea is to use the default theme implementation and not use the compatibility bodge. This would probably involve including some theme files in the source, and I would prefer a theme picker as shown https://codemirror.net/5/demo/theme.html ( this is codemirror5 but you get the idea)

### DIFF
--- a/relate/static/css/codemirror.scss
+++ b/relate/static/css/codemirror.scss
@@ -37,14 +37,15 @@
   .cmt-string    {color: #FF5500;}
 
   @media (prefers-color-scheme: dark) {
-
-    .cmt-atom      {color: #F78C6C;}
-    .cmt-comment   {color: #545454;}
-    .cmt-keyword   {color: #C792EA;}
-    .cmt-literal   {color: #FFCB6B;}
-    .cmt-number    {color: #FF5370;}
-    .cmt-operator  {color: #89DDFF;}
-    .cmt-separator {color: #FF7DE9;}
-    .cmt-string    {color: #F07178;}
+/*Converted from https://github.com/dennis84/codemirror-themes/blob/main/theme/github-dark.ts*/
+    
+    .cmt-atom      {color: #ffab70;}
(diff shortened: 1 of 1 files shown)


inducer:
Thanks! And d'oh! Just last night I *also* pushed a fix for #1084, in https://github.com/inducer/relate/commit/b3d11817008381fc9a0e630597776a3f7facca1c. Would you still prefer the Github theme? I'd be happy to bring it in if you think it's a net win.

I thought about a theme picker. (Probably we'd need two: one for dark and one for light.) That's a slightly bigger change, but manageable. If you'd like it, file an issue?

peepeetee:
> Thanks! And d'oh! Just last night I _also_ pushed a fix for #1084, in [b3d1181](https://github.com/inducer/relate/commit/b3d11817008381fc9a0e630597776a3f7facca1c). Would you still prefer the Github theme? I'd be happy to bring it in if you think it's a net win.
> 
> I thought about a theme picker. (Probably we'd need two: one for dark and one for light.) That's a slightly bigger change, but manageable. If you'd like it, file an issue?

Yee, I had to delete my browser cache but I can confirm that it is working now.

I have converted this pr to a draft, not sure how to convert it to an issue. I could close it and open a new issue if you want

inducer:
New issue sounds good if what you'd like to request is unrelated to this PR.

peepeetee:
> New issue sounds good if what you'd like to request is unrelated to this PR.

#1134

PR #1145, opened 2024-10-22
Add NaN/Inf checks into check_numpy_array_features

yslan:
For normal scenario (and provide a killer switch to turn off the check), all numpy arrays should not have NaN or Inf.


inducer:
Thanks! I agree that this is a good idea. I can't really merge it *right now*, since it might lower students' grades. But I've set a reminder for the end of the semester to merge it then (and rebuild the container image).

Also, looks like you've got a linter failure. Could you look into it?

inducer:
Thanks!

PR #1481, opened 2026-04-20
Delete the non-functional suggestions widgetry in recurring event creation

inducer:
Not much of an XSS because the source data is under instructor control.


inducer:
@copilot apply changes based on the comments in [this thread](https://github.com/inducer/relate/pull/1481#pullrequestreview-4142508468)

PR #1504, opened 2026-05-17
Switch celery task serializer to json

inducer:
x-ref: https://github.com/inducer/relate/security/advisories/GHSA-4mwh-mwv4-m252

### DIFF
--- a/relate/settings.py
+++ b/relate/settings.py
@@ -313,7 +313,7 @@ if "CELERY_BROKER_URL" not in globals():
     CELERY_BROKER_URL = "amqp://"
 
 CELERY_ACCEPT_CONTENT = ["pickle", "json"]
-CELERY_TASK_SERIALIZER = "pickle"
+CELERY_TASK_SERIALIZER = "json"
 # (pickle is buggy in django-celery-results 1.0.1)
 # https://github.com/celery/django-celery-results/issues/50
 CELERY_RESULT_SERIALIZER = "json"


emreefedogan:
Hi @inducer,

Thanks for taking the time to patch this! I reviewed the changes and noticed a critical detail regarding the `CELERY_ACCEPT_CONTENT` configuration that leaves the vulnerability partially exploitable.

While updating `CELERY_TASK_SERIALIZER = "json"` correctly forces the application to *send* new tasks as JSON, I see that the `CELERY_ACCEPT_CONTENT` list was left untouched:

```python
CELERY_ACCEPT_CONTENT = ["pickle", "json"]
```

Because `"pickle"` is still in the accepted content list, the Celery workers are still configured to *receive and deserialize* incoming pickled payloads. Since the root cause of the RCE relies on the worker deserializing untrusted data, an attacker who has access to the broker can still bypass the new serializer by sending a custom message with the `content-type` set to `application/x-python-serialize` (pickle). The worker will accept and execute this payload.

To effectively mitigate the deserialization attack vector and fully close the RCE, `"pickle"` must be entirely removed from the allowed content types:

```python
CELERY_ACCEPT_CONTENT = ["json"]
```

This ensures the worker will instantly reject any pickled data it receives from the broker, completely neutralizing the threat.

By the way, thank you for requesting the CVE. 

Best regards,

inducer:
Thanks for pointing that out. I should not have missed that.

https://github.com/inducer/relate/commit/d66ba5659b459bf1ba56b7109b5f9ecf197cbefb (pushed to `main` by accident, not via a PR as I should have)

emreefedogan:
Hi @inducer,

Thank you for the quick turnaround! I've checked the latest commit (d66ba56), and it looks solid. Removing pickle entirely from the allowed content types effectively neutralizes the deserialization attack vector and fully closes the RCE.

Do you have an estimated timeline for when you plan to cut a new release and publish the Security Advisory (GHSA-4mwh-mwv4-m252)? Please let me know if there is anything else you need from my side to help wrap up this process.

Thanks again for the smooth collaboration.

Best regards,

inducer:
I've published the advisory last week. Relate does not currently have a formal release process. ("Just get it from Git!" sorry, world) So there isn't really a release that needs cutting.

